PR #4, opened 2025-02-26
Add release/deploy CLI

yuriy-arabskyy:


src/qubx/core/loggers.py
```diff
@@ -84,6 +84,10 @@ def get_portfolio(self, as_plain_dataframe=True) -> pd.DataFrame:
                     },
                     axis=1,
                 )
+                # We want to convert the value to just price * quantity
+                # in reality value of perps is just the unrealized pnl but
+                # it's not important after simulation for metric calculations
+                pi["Value"] = pi["Pos"] * pi["Price"] + pi["Value"]
```
yuriy-arabskyy, line 90: after my old updates to the definition of Value, Value is defined as the unrealized pnl for perp positions. This is how the exchanges handle this and how it's needed for correct margin calculations. But because of this, our metric calculations didn't work correctly and leverage was not working. So now I export into the portfolio log the standard definition of Value to make the metrics work correctly as before.

src/qubx/__init__.py
```diff
@@ -5,13 +5,33 @@
...
 from loguru import logger
 
-from qubx.core.lookups import FeesLookup, GlobalLookup, InstrumentsLookup
-from qubx.utils import runtime_env, set_mpl_theme
-from qubx.utils.misc import install_pyx_recompiler_for_dev
-
 # - TODO: import some main methods from packages
 
 
+def runtime_env():
```
yuriy-arabskyy, line 11: I moved runtime_env in here to remove the qubx.utils import from the main init, because this slows down the initialization by a lot when running qubx CLI commands

dmarienko (commented):
Ok, looks good !

dmarienko (approved):

PR #12, opened 2025-03-04
Mde/dev -> dev

dmarienko:
1. Account processor `update_position_price` should handle trades, quotes 
2.OME should handle quotes properly so that limit orders between spread can get filled properly. Also OME should handle trades.
3. Add special order flag to execute at specific price (it has to be between the current quote and previous quote)


yuriy-arabskyy:
Looks good!

yuriy-arabskyy (approved):

PR #20, opened 2025-03-12
Mde/deploy fixes

dmarienko:


yuriy-arabskyy:
It looks good, the tests are passing so let's merge :) Maybe we can already remove the code that release by strategy name and not config

yuriy-arabskyy (approved):

PR #23, opened 2025-03-14
Live fixes

yuriy-arabskyy:
1. Rewrite orderbook creation from ccxt response
2. Implement tardis machine reader
3. Add custom client id for orders
4. Add custom maker,taker support for fees
5. Add realized pnl to total capital calculation
6. Wrap process data in a try except
7. Move integration and e2e tests to corresponding folders
8. Rework emitter design (add time provider, add optional instrument argument)


src/qubx/notifications/slack.py
dmarienko, line 1: I'm bit concerned about latency when we do this rest slack call, but as I see we use it only in start/stop/error scenarios, right ?

src/qubx/core/lookups.py
```diff
@@ -372,12 +372,27 @@ def refresh(self):
...
+    def find(self, exchange: str, spec: str | None) -> TransactionCostsCalculator:
         if spec is None:
             return ZERO_COSTS
+
         key = f"{exchange}_{spec}"
+
+        # - check if spec is of type maker=...,taker=...
+        # Check if spec is in the format maker=X,taker=Y
+        maker_taker_pattern = re.compile(r"maker=(-?[0-9.]+),taker=(-?[0-9.]+)")
+        match = maker_taker_pattern.match(spec)
```
dmarienko, line 384: I'd propose to use bit extended regexp:
re.compile(r"maker=(-?[0-9.]+)[,\ ]+taker=(-?[0-9.]+)")

it'd allow us to use less tight syntax in fees spec: `maker=1 taker=2` or `maker=1,taker=2` etc

src/qubx/backtester/account.py
```diff
@@ -154,7 +154,9 @@ def _process_new_data(self, instrument: Instrument, data: Quote | OrderBook | Tr
             return
         for r in ome.process_market_data(data):
             if r.exec is not None:
-                self.order_to_instrument.pop(r.order.id)
+                # TODO: why do we pop here if it is popped later after send
+                if r.order.id in self.order_to_instrument:
+                    self.order_to_instrument.pop(r.order.id)
```
dmarienko, line 158: theoretically someone can try to try cancel order in on_execution callback (it'd be called directly in send call) - so it's removed here

src/qubx/notifications/slack.py
yuriy-arabskyy: but I agree, there are also signal exporters that can slow processing down. I'll add threads inside of these things

yuriy-arabskyy: and emitters

src/qubx/core/lookups.py
yuriy-arabskyy: fixed

PR #26, opened 2025-03-21
Implement async trading and on error handler

yuriy-arabskyy:


dmarienko (commented):
Looks fine as for me !

PR #32, opened 2025-03-28
feat: redesign of simulated exchange

dmarienko:
- now all code related to orders simulation is moved to BasicSimulatedExchange class (OME is there too and it just one of possible implementations)
- it's possible to create any custom exchanges simulators (like for stocks, futures) which could emulate some particular features: custom order flags, rejections, latencies etc

### DIFF
--- a/src/qubx/backtester/simulated_exchange.py
+++ b/src/qubx/backtester/simulated_exchange.py
@@ -14,15 +14,85 @@ from qubx.core.basics import (
 from qubx.core.series import Bar, OrderBook, Quote, Trade, TradeArray
 
 
-class BasicSimulatedExchange:
+class ISimulatedExchange:
     """
-    Basic emulation of generic exchange.
+    Generic interface for simulated exchange.
     """
 
     exchange_id: str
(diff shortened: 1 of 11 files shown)


src/qubx/backtester/simulated_exchange.py
```diff
@@ -0,0 +1,193 @@
...
+    ITimeProvider,
+    Order,
+    Timestamped,
+    TransactionCostsCalculator,
+    dt_64,
+)
+from qubx.core.series import Bar, OrderBook, Quote, Trade, TradeArray
+
+
+class BasicSimulatedExchange:
```
yuriy-arabskyy: maybe you could define an interface for the simulated exchange? For other possible simulated exchanges to implement.

yuriy-arabskyy:
Does it fix now the issue that we have in paper mode?

yuriy-arabskyy (approved):

src/qubx/backtester/simulated_exchange.py
dmarienko: It's done

PR #38, opened 2025-04-15
CsvFileLogsWriter: added autoclosing after write

bogdanKaftanatiy:


src/qubx/core/loggers.py
```diff
@@ -155,25 +149,28 @@ def _do_write(self, log_type, data):
                     w.writerows(self._values(data))
 
             case "portfolio":
-                if self._hdr_pfl:
-                    self._pfl_writer.writerow(self._header(data[0]))
-                    self._hdr_pfl = False
-                self._pfl_writer.writerows(self._values(data))
-                self._pfl_file_.flush()
+                with open(self._portfolio_file_path, "w", newline="") as f:
```
yuriy-arabskyy, line 152: you should definitely open in "a" append mode and not write mode, because otherwise you overwrite the file every time. Also I'm not sure if it's ok to open and close the file on every execution and signal. @dmarienko what do you think?

The reason for this PR is that these logs are written to google bucket and without closing the file, they don't get synced on the bucket.

@bogdanKaftanatiy I'm thinking maybe it's better to just implement a MongoLogsWriter and have it done properly

bogdanKaftanatiy: Yeah, makes sense.
I'm going to close this PR, and will add `MongoLogsWriter` and `MongoStateRestorer` to handle live qubx in cluster

bogdanKaftanatiy:
Closing this PR, as it is suboptimal for live execution in cluster. Will proceed with mongo based implementation

PR #39, opened 2025-04-16
Minor fixes

yuriy-arabskyy:


src/qubx/connectors/ccxt/data.py
```diff
@@ -616,9 +616,10 @@ async def watch_quote(instruments: list[Instrument]):
             for exch_symbol, ccxt_ticker in ccxt_tickers.items():  # type: ignore
                 instrument = ccxt_find_instrument(exch_symbol, self._exchange, _symbol_to_instrument)
                 quote = ccxt_convert_ticker(ccxt_ticker)
-                self._health_monitor.record_data_arrival(sub_type, dt_64(quote.time, "ns"))
-                self._last_quotes[instrument] = quote
-                channel.send((instrument, sub_type, quote, False))
+                if self._last_quotes[instrument] is None or quote.time > self._last_quotes[instrument].time:
```
dmarienko, line 619: what if we receive different quotes with same time ? It happens sometimes so may be it worth to use >= there ?

dmarienko (commented):
Looks good as for me !

src/qubx/connectors/ccxt/data.py
yuriy-arabskyy: I'll change later, I think it's very low probability

PR #42, opened 2025-04-17
MongoDB logger and state restorer

bogdanKaftanatiy:
Extended qubx logging with mongodb logger and state restorers implementation. 
Tested in unit tests and live paper mode for simple ObiTraderStrategy


src/qubx/restorers/state.py
```diff
@@ -112,3 +113,102 @@ def restore_state(self) -> RestoredState:
...
+        self.strategy_id = strategy_id
+
+        self.client = MongoClient(mongo_uri)
+        self.collection = self.client[db_name][collection_name]
+
+        # Create individual restorers
+        self.position_restorer = MongoDBPositionRestorer(
+            bot_id=bot_id,
+            strategy_id=strategy_id,
+            mongo_uri=mongo_uri,
```
yuriy-arabskyy: create the mongo client just once for the restorers and pass the same client in each of the restorers

src/qubx/restorers/state.py
```diff
@@ -112,3 +113,102 @@ def restore_state(self) -> RestoredState:
...
+            db_name=db_name,
+            collection_name=collection_name,
+        )
+
+        self.signal_restorer = MongoDBSignalRestorer(
+            bot_id=bot_id,
```
yuriy-arabskyy: I don't like the introduction of bot_id and strategy_id in the framework. These are coming from another system, instead of imposing this constraint on qubx, maybe it's better to just use a unified strategy_name everywhere without complications. And then if you have bot_id and strategy_id, you can encode it into a strategy_name when starting the strategy. What do you think?

src/qubx/core/loggers.py
```diff
@@ -201,6 +203,71 @@ def close(self):
...
+
+    def __init__(
+        self,
+        account_id: str,
+        strategy_id: str,
+        run_id: str,
+        bot_id: str,
+        mongo_uri: str = "mongodb://localhost:27017/",
+        db_name: str = "default_logs_db",
+        collection_name: str = "qubx_logs",
```
yuriy-arabskyy: why not separate positions, signals, balances, etc into separate collections and not mix them up? It will be later easier to manage this I think

src/qubx/core/loggers.py
```diff
@@ -201,6 +203,71 @@ def close(self):
         self.pool.join()
 
 
+class MongoDBLogsWriter(LogsWriter):
```
yuriy-arabskyy: I think it makes sense to refactor into a separate package `loggers` and then have a separate file for csv, mongo, etc. Just like restorers and other things. And in qubx/core/loggers we could just keep the interfaces, otherwise this file will just grow

dmarienko: Also it'd be really cool if we could handle them as optional dependencies, not only mongo but questdb, tardis - all specific stuff etc.
Something like `pip install qubx[extras]` etc to reduce size and initialization time.

src/qubx/restorers/factory.py
```diff
@@ -8,33 +8,37 @@
...
     "CsvBalanceRestorer": CsvBalanceRestorer,
+    "MongoDBBalanceRestorer": MongoDBBalanceRestorer,
 }
 
 
 # Registry of state restorer types
 STATE_RESTORER_REGISTRY: dict[str, Type[IStateRestorer]] = {
     "CsvStateRestorer": CsvStateRestorer,
+    "MongoDBStateRestorer": MongoDBStateRestorer,
 }
```
dmarienko, line 43: Btw, what if we move this registry lookup into some resource file ? It'd allow to attach custom restorers if need.

src/qubx/core/loggers.py
bogdanKaftanatiy: hm, yeah, sounds interesting, but not exactly sure how to do it yet. I think we can add later, do you agree?

src/qubx/restorers/factory.py
bogdanKaftanatiy: sorry, not following your idea here. Could you elaborate?

dmarienko: ok, something like having resource file in qubx package where we could put this map:
like src/qubx/resources/registry.yml
```
positions:
    csv: qubx.restorers.positions.CsvStateRestorer,
    mongo: qubx.restorers.positions.MongoDbrestorer,
    questdb: ......
    
signals:
    csv: .....  
    
etc
```
Then we load it in factory and also look if there is registry.yml file in .qubx/ or in local directory, if it is we override default by custom setups.

Actually we could use this registry for all custom factories we use in Qubx (like positions loggers etc).

yuriy-arabskyy: I like this and we can do this, but maybe let's Bogdan finish for now this first iteration just to get this running

src/qubx/core/loggers.py
yuriy-arabskyy: of course

yuriy-arabskyy (approved):
nice! I think it looks good and ok to merge, @dmarienko want to take a final look too?

dmarienko:
Yes, looks good!

PR #46, opened 2025-05-14
Exporters: added leverages initialization for redis exporter

bogdanKaftanatiy:
After restart previous leverages information for redis restorer is empty, leading to first incorrect position change export. Added leverages fetching during initialization


yuriy-arabskyy (approved):
looks good, just make sure that it works in live

PR #59, opened 2025-06-11
Dm/lookup

dmarienko:
New lookup interface design.
Now it supports reading .qubx/setting.ini config where we can specify lookups settings.
for example: 

this settings.ini would configure using mongo database for reading instruments lookups data and 
file for fees lookup.
```
[instrument-lookup]
type=mongo
mongo_url=mongodb://localhost:27017/
reload_interval=12h   # how frequent refresh actual lookup data
#type=file
#query_exchanges=true


[fees-lookup]
type=file
```  
If config file doesn't exist it uses just plain files lookups (as we used before).

Also now all delivery futures instruments have expiration date in it's symbol convention:
so Binance.UM BTCUSDT future expired at Sep'25 looks like that: BINANCE.UM:FUTURE:BTCUSDT.20250914 etc

This gives us clean way to specify exact names in simulations and live trading and trade different calendars for same symbol.

Also it fixes delisted date parsing (now it takes it in account from tardis data)


bogdanKaftanatiy (approved):
LGTM

src/qubx/core/basics.py
```diff
@@ -879,3 +889,89 @@ class RestoredState:
     balances: dict[str, AssetBalance]
     instrument_to_target_positions: dict[Instrument, list[TargetPosition]]
     positions: dict[Instrument, Position]
+
+
+class InstrumentsLookup:
+    def get_lookup(self) -> dict[str, Instrument]: ...
+
+    def get_market_type(self, exchange: str) -> MarketType | None: ...
```
yuriy-arabskyy: an exchange can have multiple market types, maybe it doesn't make sense to have it like this? Because we can trade from the same exchange account margin/perps/futures

src/qubx/core/basics.py
```diff
@@ -879,3 +889,89 @@ class RestoredState:
...
+            if (
+                (i.exchange == exchange)
+                and (i.symbol == symbol)
+                and (market_type is None or i.market_type == market_type)
+            ):
+                return i
+
+        return None
+
+    def find_instruments(
```
yuriy-arabskyy, line 926: do we want to support the as_of optional date for finding instruments? So that we get active for that period

src/qubx/core/lookups.py
```diff
@@ -550,36 +477,139 @@ def _convert_instruments_metadata_to_qubx(data: list[dict]):
...
+class LookupsManager(InstrumentsLookup, FeesLookup):
+    _i_lookup: InstrumentsLookup
+    _t_lookup: FeesLookup
+
+    def __new__(cls):
+        if not hasattr(cls, "instance"):
+            cls.instance = super(LookupsManager, cls).__new__(cls)
+
+            # - try to load settings
+            parser = configparser.ConfigParser()
```
yuriy-arabskyy, line 449: as I mentioned in the chat, would be nicer to have a more unified approach to config management. Similar to dvc for instance, if we take the core config management logic from there and use pydantic base settings for management. This is very nice, because then we can have configs in json files and can also override with env variables and this is very handy when deploying live

src/qubx/core/lookups.py
```diff
@@ -550,36 +477,139 @@ def _convert_instruments_metadata_to_qubx(data: list[dict]):
...
+class InstrumentsLookupMongo(InstrumentsLookup):
+    _MONGO_DB_BASE_NAME = "metadata"
+    _MONGO_DB_TABLE_NAME = "instruments"
+
+    _lookup: dict[str, Instrument]
+    _mongo_url: str
+
+    def __init__(self, mongo_url: str = "mongodb://localhost:27017/"):
+        self._mongo_url = mongo_url
+        self.load()
```
yuriy-arabskyy, line 416: how do we make sure that in live this is updated regularly? because as it is now, it will have a fixed list at start and we need to manually call the load() from the strategy code. I think this logic should be baked in somewhere so that the strategy doesn't need to know about these details

src/qubx/core/basics.py
dmarienko: This is some old useless code - I've removed it at all

src/qubx/core/lookups.py
dmarienko: Yes, agree, but I'd rather do some research on what's best framework for configs management. I really don't like ini approach.
May be we could use something like https://www.dynaconf.com/ that supports a lot of options.

src/qubx/core/lookups.py
dmarienko: We could just add another mongo lookup reader that would just reload lookups in background (or even on every request).

src/qubx/core/basics.py
dmarienko: Yes, we'd need unified query method for requesting instruments, including all these options etc. But not sure it's super urgent rn.

yuriy-arabskyy: it's useful now, because for the ingestion code we need to do requests to get active instruments for a particular point in time to figure out what to ingest. But we can do that purely in the datavault logic for now to keep this interface simple

src/qubx/core/lookups.py
yuriy-arabskyy: I agree, also really don't like ini files. I'll read a bit more. But please check what you think about the approach that I used https://github.com/xLydianSoftware/datavault/tree/main/projects/dvc/src/dvc/configs , it's the first time I use this design for configs, so not sure myself.

src/qubx/core/lookups.py
dmarienko: ok, added option for reloading mongo data.
So it'd try to update lookup info if at least reload_interval time is passed from previous query.

src/qubx/core/basics.py
dmarienko: Ok, I see it's quite easy to add there - let's keep all logic in qubx then.

dmarienko: added parameter for this case

PR #60, opened 2025-06-17
fix: finalize OHLC data for all instruments in the universe on trigge…

dmarienko:


src/qubx/core/mixins/processing.py
```diff
@@ -198,12 +198,32 @@ def __process_data(self, instrument: Instrument, d_type: str, data: Any, is_hist
...
                     signals = self._wrap_signal_list(self._strategy.on_market_data(self._context, event))
 
-            if isinstance(event, TriggerEvent) or (isinstance(event, MarketEvent) and event.is_trigger):
-                _trigger_event = event.to_trigger() if isinstance(event, MarketEvent) else event
+            if _is_trigger_ev or (_is_market_ev and event.is_trigger):
+                _trigger_event = event.to_trigger() if _is_market_ev else event
+
+                # FIXME: (2025-06-17) we need to refactor this to avoid doing it here !!!
+                # - on trigger event we need to be sure that all instruments have finalized OHLC data
+                if not self._is_simulation:
```
yuriy-arabskyy, line 214: it looks good, but now I think that this should be also applied in simulation, because simulations have the same issue. Do you agree?

yuriy-arabskyy:
I am merging for now, but let's discuss the simulation mode.

PR #63, opened 2025-06-26
Dm/targets restoring

dmarienko:
- Split targets / signals


src/qubx/core/mixins/processing.py
```diff
@@ -283,60 +393,33 @@ def __invoke_on_fit(self) -> None:
...
-            signals = [signals]
-        elif signals is None:
-            return []
-
-        for signal in signals:
-            # set strategy group name if not set
-            if not signal.group:
-                signal.group = self._strategy_name
+                    # - new position will be non-zero
+                    if abs(_existing_pos + t.target_position_size) > _instr.min_size:
```
yuriy-arabskyy: why do you add target to existing position? target is target, so current position does not influence the target position

src/qubx/core/mixins/processing.py
```diff
@@ -694,3 +791,10 @@ def _log_state_mismatch(self) -> None:
                 f"Warmup: [Position: {warmup_pos_info}, {warmup_ord_info}] | "
                 f"Current: [Position: {current_pos_info}, {current_ord_info}]"
             )
+
+    def get_active_targets(self) -> dict[Instrument, TargetPosition]:
+        return self._active_targets
```
yuriy-arabskyy, line 789: q: it's not yet completely clear how to interpret the concept of active target positions. What happens to the active target position if the strategy does a trade instead of emitting a signal?

src/qubx/core/mixins/processing.py
```diff
@@ -119,6 +130,13 @@ def __init__(
         self._data_ready_start_time = None
         self._last_data_ready_log_time = None
         self._all_instruments_ready_logged = False
+        self._emitted_signals = []
+
+        # - special tracker for post-warmup initialization signals
+        self._init_stage_position_tracker = _InitializationStageTracker()
+        self._instruments_in_init_stage = set()
+        self._active_targets = {}
+        self._trackers_map = {}
```
yuriy-arabskyy: I think the trackers map is not needed here, it's not used anyway

src/qubx/core/context.py
```diff
@@ -522,16 +525,28 @@ def get_event_schedule(self, event_id: str) -> str | None:
     def is_fitted(self) -> bool:
         return self._processing_manager.is_fitted()
 
+    def get_active_targets(self) -> dict[Instrument, list[TargetPosition]]:
+        return self._processing_manager.get_active_targets()
```
yuriy-arabskyy, line 531: why do you get the active targets and not the actual positions? I'm just worried what if target position is not in sync with the current position

yuriy-arabskyy: but I guess it's useful because TP and SL is connected to it directly

src/qubx/core/mixins/processing.py
yuriy-arabskyy: @dmarienko this is the only logical mistake I see so far, it should just be `abs(t.target_position_size) > _instr.min_size` right?

src/qubx/core/mixins/processing.py
dmarienko: Well, we must add target positions generating for any trading interaction in future. So any trade_xxx() should change target position (or may be we can so that using signals).

src/qubx/core/mixins/processing.py
dmarienko: Good, catch ! Will drop it.

src/qubx/core/context.py
dmarienko: Yes, we have TP (as some state) any time and can store / restore it.

src/qubx/core/mixins/processing.py
dmarienko: Yep, looks like a mistake, good one ! This would be cleared in handling executions report anyway but it's definitely mistake.

PR #69, opened 2025-07-17
Fix order not found warning

yuriy-arabskyy:


src/qubx/backtester/broker.py
```diff
@@ -63,8 +64,12 @@ def send_order_async(
         self.send_order(instrument, order_side, order_type, amount, price, client_id, time_in_force, **optional)
 
     def cancel_order(self, order_id: str) -> Order | None:
-        self._send_execution_report(order_update := self._exchange.cancel_order(order_id))
-        return order_update.order if order_update is not None else None
+        try:
+            self._send_execution_report(order_update := self._exchange.cancel_order(order_id))
+            return order_update.order if order_update is not None else None
+        except OrderNotFound:
+            # Order was already cancelled or doesn't exist
```
dmarienko, line 72: Shouldn't we add debug/warning message to log here ?

dmarienko (approved):

src/qubx/backtester/broker.py
yuriy-arabskyy: true

PR #70, opened 2025-07-17
cctx ws connection aliveness check and ws reconenction

bogdanKaftanatiy:


src/qubx/connectors/ccxt/data.py
```diff
@@ -89,6 +90,11 @@ def __init__(
         self._is_sub_name_enabled = defaultdict(lambda: False)
         self._sub_connection_ready = defaultdict(lambda: False)
         self._symbol_to_instrument = {}
+        
+        # Simple health monitoring
+        self._last_data_time: Dict[str, float] = {}
+        self._connection_timeout = 120.0  # 2 minutes without data = stale connection
```
dmarienko, line 96: I think it's better not to use 'magic' constants directly in the code. What's about making it named constant ?

src/qubx/connectors/ccxt/data.py
```diff
@@ -221,6 +230,15 @@ async def _get():
         return _arr
 
     def close(self):
+        # Cancel health monitor
+        if hasattr(self, '_health_task') and self._health_task is not None:
```
dmarienko, line 234: Do we really need to check if object has _health_task if it's initialized in init ? 
Also may be it's good to declare _health_task as filed in class as _health_task: xxx | None  ?

src/qubx/connectors/ccxt/data.py
```diff
@@ -231,6 +249,70 @@ def close(self):
...
+        """Start simple health monitoring task"""
+        if not hasattr(self, '_health_task') or self._health_task is None:
+            self._health_task = self._loop.submit(self._health_monitor_loop())
+
+    async def _health_monitor_loop(self):
+        """Simple health monitoring loop"""
+        while self.channel.control.is_set():
+            try:
+                await asyncio.sleep(30)  # Check every 30 seconds
+                current_time = time.time()
```
dmarienko, line 262: It's better to avoid using system time, let's use time_provider instead to keep code consistent.

PR #71, opened 2025-07-18
Implement stale data detector

yuriy-arabskyy:


src/qubx/core/stale_data_detector.py
```diff
@@ -0,0 +1,170 @@
...
+            # Calculate how many bars we need to check based on timeframe
+            timeframe_seconds = self._cache.default_timeframe / td_64(1, "s")
+            detection_period_seconds = self._detection_period / td_64(1, "s")
+            bars_to_check = max(int(detection_period_seconds / timeframe_seconds), self._min_bars_required)
+            
+            # Check the last N bars
+            bars_checked = 0
+            prev_bar = None
+            
+            for i in range(min(bars_to_check, len(ohlcv))):
```
dmarienko, line 78: What if we memorize last checked bar and start from it to reduce loop ?

dmarienko (commented):
Looks good, but I'm bit concerned about speed of this additional check - may be we could move it into cython ?

src/qubx/core/stale_data_detector.py
yuriy-arabskyy: yes, that'a a good idea with memorizing last checked bar, will do. I had to do this stale check because there are issues with some data of smaller market cap coins when I check for 200 or so

PR #74, opened 2025-07-27
data gathering: additional subscriptions

bogdanKaftanatiy:


yuriy-arabskyy:
@claude review the PR

src/qubx/connectors/ccxt/data.py
```diff
@@ -819,5 +874,126 @@ async def un_watch_funding_rates():
...
+        def should_poll_now():
+            """Check if we should poll now (every 5 minutes at boundaries)"""
+            now = datetime.datetime.now(datetime.timezone.utc)
+            minute_mod = now.minute % 5
+            is_boundary = minute_mod == 0
+            is_early_in_boundary = now.second < 30
+            should_poll = is_boundary and is_early_in_boundary
+            return should_poll
+        
+        async def poll_open_interest_once():
```
yuriy-arabskyy: why do you poll via REST open interest? didn't we see with you that markPrice subscription contains this information?

src/qubx/connectors/ccxt/data.py
```diff
@@ -789,22 +820,46 @@ async def _subscribe_funding_rate(
         sub_type: str,
         channel: CtrlChannel,
     ):
+        # Add a small delay if this is a subscription update to prevent race conditions
+        # This gives time for any pending watch_mark_prices calls to complete
+        if "funding_rate" in name.lower():
+            logger.debug(f"<yellow>{self._exchange_id}</yellow> Adding 1s delay to prevent funding rate subscription race condition...")
```
yuriy-arabskyy: are you sure you need this? Also why do you check "funding_rate" in the name if you are already in the subscribe funding rate method

src/qubx/connectors/ccxt/data.py
```diff
@@ -789,22 +820,46 @@ async def _subscribe_funding_rate(
...
-            channel.send((None, sub_type, instrument_to_funding_rate, False))
+            try:
+                # Get the symbols from subscribed instruments
+                subscribed_instruments = self.get_subscribed_instruments(sub_type)
+                symbols = [instr.symbol for instr in subscribed_instruments] if subscribed_instruments else []
+                
+                funding_rates = await self._exchange.watch_funding_rates(symbols)  # Pass symbols parameter
+                
+                # WebSocket streams return one symbol per message - this is normal behavior
+                symbol_count = len(funding_rates)
```
yuriy-arabskyy: clean up unused variables

src/qubx/core/series.pxd
```diff
@@ -59,8 +59,9 @@ cdef class Bar:
     cdef public double close
     cdef public double volume          # total volume (in quote asset)
```
yuriy-arabskyy: if you're changing this, could you also add volume_quote and bought_volume_quote? also update the comments are wrong I think, because volume and bought_volume are in base asset

src/qubx/core/series.pyx
```diff
@@ -746,34 +746,37 @@ cdef class Quote:
...
             }
         return {
             'timestamp': np.datetime64(self.time, 'ns'), 
             'open': self.open, 'high': self.high, 'low': self.low, 'close': self.close, 
             'volume': self.volume,
             'bought_volume': self.bought_volume,
+            'trade_count': self.trade_count,
         }
 
     def __repr__(self):
```
yuriy-arabskyy, line 790: can you add also time_to_str for the string representation of the bar? small change but will be useful for debugging

src/qubx/connectors/ccxt/data.py
```diff
@@ -213,7 +216,9 @@ async def _get():
         _arr = []
         for oh in res:  # type: ignore
             _arr.append(
-                Bar(oh[0] * 1_000_000, oh[1], oh[2], oh[3], oh[4], oh[6], oh[7])
+                Bar(oh[0] * 1_000_000, oh[1], oh[2], oh[3], oh[4], oh[6], oh[7], int(oh[8]))
+                if len(oh) > 8
```
yuriy-arabskyy: could you also figure out how to get volume quote, bought volume, bought volume quote? We need this for another strategy

src/qubx/connectors/ccxt/data.py
```diff
@@ -268,7 +273,28 @@ def _subscribe(
             self._pending_subscriptions.pop(_sub_type, None)
             self._sub_connection_ready.pop(sub_type, None)
 
+            # Explicitly disable the old subscription to make it exit its loop
+            if old_sub_info:
```
yuriy-arabskyy: why do you check if this is not None if we constructed this object just a couple of lines above in the same code?

src/qubx/connectors/ccxt/data.py
bogdanKaftanatiy: binance mark price ws stream doesn't provide open interest info
https://developers.binance.com/docs/derivatives/usds-margined-futures/websocket-market-streams/Mark-Price-Stream

src/qubx/connectors/ccxt/data.py
dmarienko: I'd propose to make some utility method that take care about converting from array to Bar with all this length comparison logic.

src/qubx/connectors/ccxt/data.py
```diff
@@ -819,5 +867,126 @@ async def un_watch_funding_rates():
...
+        """
+        Subscribe to open interest data using REST API polling.
+        Binance doesn't provide open interest via WebSocket, so we poll every 5 minutes at exact boundaries.
+        Polls at 00, 05, 10, 15, 20, etc. minutes of each hour.
+        """
+        
+        def get_current_5min_boundary():
+            """Get the current 5-minute boundary timestamp for data timestamping"""
+            now = datetime.datetime.now(datetime.timezone.utc)
+            current_minute = (now.minute // 5) * 5
```
dmarienko: It'd be better to avoid using any datetime.now() calls here and use self.time_provider instead.

src/qubx/connectors/ccxt/data.py
```diff
@@ -819,5 +867,126 @@ async def un_watch_funding_rates():
...
+                        
+                    except CcxtSymbolNotRecognized as e:
+                        logger.warning(f"<yellow>{self._exchange_id}</yellow> Symbol not recognized: {symbol}")
+                        continue
+                    except Exception as e:
+                        logger.error(f"<yellow>{self._exchange_id}</yellow> Error fetching open interest for {symbol}: {type(e).__name__}: {e}")
+                        continue
+                
+                # After successful poll, sleep longer until next check (but still cancellation-aware)
+                await cancellation_aware_sleep(100)
```
dmarienko: Also let's move all timeouts as named constants like CANCEL_TIMEOUT_SLEEP_MS instead of using magic number in the code

PR #76, opened 2025-07-28
Refactor ccxt data provider

yuriy-arabskyy:


src/qubx/connectors/ccxt/connection_manager.py
```diff
@@ -0,0 +1,284 @@
...
+    def set_subscription_manager(self, subscription_manager: SubscriptionManager) -> None:
+        """Set the subscription manager for state coordination."""
+        self._subscription_manager = subscription_manager
+    
+    async def listen_to_stream(
+        self,
+        subscriber: Callable[[], Awaitable[None]],
+        exchange: Exchange,
+        channel: CtrlChannel,
+        stream_name: str,
```
bogdanKaftanatiy, line 60: in most places we use stream name as the main unique identificator of active ws connection. But what I noticed, that is possible to have several active ws connections for the same stream name at the same time. 
For example for some reason we didn't stop old connection, but opened a new one. In such case we are going override connection references
So I think extra validation would be good to add, so we check if old connection for this stream name exists

bogdanKaftanatiy: ah, okay, I see that there is resubscription_cleanup flow added for that, all good

PR #87, opened 2025-08-22
ccxt reconnection

bogdanKaftanatiy:
added ccxt exchange wrapper that allows to perform full reconnection


src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,163 @@
...
+            params = self._factory_params.copy()
+            params.pop('enable_stability_manager', None)
+            exchange = get_ccxt_exchange(**params)
+            logger.debug(f"Created new {self._exchange_name} exchange instance")
+            return exchange
+        except Exception as e:
+            logger.error(f"Failed to create {self._exchange_name} exchange: {e}")
+            raise RuntimeError(f"Failed to create {self._exchange_name} exchange: {e}") from e
+
+    def force_recreation_on_stall(self) -> bool:
```
yuriy-arabskyy: do you mean stale?

src/qubx/health/base.py
```diff
@@ -111,6 +114,19 @@ def stop(self) -> None:
         """Stop the health metrics monitor."""
         pass
 
+    def register_exchange_manager(
```
yuriy-arabskyy: I don't like this design, it is adding dependency in the base health manager to a concrete ccxt implementation, that is not ok

src/qubx/connectors/ccxt/factory.py
```diff
@@ -62,7 +92,18 @@ def get_ccxt_exchange(
     if use_testnet:
         ccxt_exchange.set_sandbox_mode(True)
 
-    return ccxt_exchange
+    # Wrap in ExchangeManager if stability management enabled
+    if enable_stability_manager:
```
yuriy-arabskyy: why do you need backwards compatibility?

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,163 @@
...
+        
+        if time_since_reset >= reset_interval_seconds and self._recreation_count > 0:
+            logger.info(f"Resetting recreation count for {self._exchange_name} (was {self._recreation_count})")
+            self._recreation_count = 0
+            self._last_successful_reset = current_time
+
+    # === Transparent Delegation === 
+    # All CCXT exchange methods/properties are delegated to underlying exchange
+    
+    def __getattr__(self, name: str) -> Any:
```
yuriy-arabskyy: This will lead to all dependant code not being able to see the actual methods of the exchange because they are now dynamically added which is a bad practice. Why not expose the exchange via a @property exchange and update the code in ccxt connector to get the exchange manager object and then access the exchange via the exchange property instead of adding methods dynamically

src/qubx/connectors/ccxt/data.py
```diff
@@ -76,7 +82,7 @@ def __init__(
         # Data type handler factory for clean separation of data processing logic
         self._data_type_handler_factory = DataTypeHandlerFactory(
             data_provider=self,
-            exchange=self._exchange,
+            exchange=self._exchange_manager.exchange,
```
yuriy-arabskyy: I think when you pass the exchange like this, the handlers will always have a reference to the old exchange, so I think you need to refactor the handler to receive the exchange manager here

src/qubx/connectors/ccxt/data.py
```diff
@@ -92,40 +98,32 @@ def __init__(
...
+    
+    def notify_data_arrival(self, event_type: str, event_time: dt_64) -> None:
+        """Notify all registered listeners about data arrival.
+        
+        Args:
+            event_type: Type of data event (e.g., "ohlcv:BTC/USDT:1m")
+            event_time: Timestamp of the data event
         """
-        asyncio_loop = self._exchange.asyncio_loop
+        for listener in self._data_arrival_listeners:
```
yuriy-arabskyy, line 117: why do you need a list of listeners and not just one listener?

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,410 @@
...
+    - Atomic exchange transitions during recreation
+    - Background monitoring thread for stall detection
+    """
+    
+    _exchange: cxp.Exchange  # Type hint that this is always a valid exchange
+
+    def __init__(
+        self,
+        exchange_name: str,
+        factory_params: Dict[str, Any],
```
yuriy-arabskyy: can you try to use in the future dict, list, etc instead of Dict, List from typing. This is a minor thing but we are trying to use the newer standard

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,410 @@
...
+    
+    def on_data_arrival(self, event_type: str, event_time: dt_64) -> None:
+        """Record data arrival for stall detection.
+        
+        Args:
+            event_type: Type of data event (e.g., "ohlcv", "trade", "orderbook")
+            event_time: Timestamp of the data event (unused for stall detection)
+        """
+        current_timestamp = time.time()
+        with self._data_lock:
```
yuriy-arabskyy, line 210: do you need a lock here if all data logic is single threaded asynchronous code

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,410 @@
...
+    
+    def _extract_ohlc_timeframe(self, event_type: str) -> Optional[str]:
+        """Extract timeframe from OHLC event type like 'ohlc(1m)' -> '1m'."""
+        if event_type.startswith('ohlc(') and event_type.endswith(')'):
+            return event_type[5:-1]  # Simple slice: ohlc(1m) -> 1m
+        return None
+
+    def _timeframe_to_seconds(self, timeframe: str) -> int:
+        """Convert timeframe string to seconds using pandas.Timedelta."""
+        import pandas as pd
```
yuriy-arabskyy: import at the top

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,410 @@
...
+        return int(pd.Timedelta(timeframe).total_seconds())
+
+    def _get_stall_threshold(self, event_type: str) -> float:
```
yuriy-arabskyy, line 293: *stale

src/qubx/connectors/ccxt/data.py
bogdanKaftanatiy: There are two listeners already: health monitor and exchange manager (see init method)

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,410 @@
...
+        
+        # Use provided exchange or create new one
+        if initial_exchange:
+            self._exchange = initial_exchange
+            # Setup exception handler on provided exchange
+            self._setup_ccxt_exception_handler(self._exchange)
+        else:
+            self._exchange = self._create_exchange()
+
+    def _create_exchange(self) -> cxp.Exchange:
```
yuriy-arabskyy, line 103: why did you move so much of the ccxt factory logic in this class instead of calling the get_ccxt_exchange from the factory?

src/qubx/connectors/ccxt/exchange_manager.py
bogdanKaftanatiy: same `_last_data_times` is accessed in stalls checking, so it could be updated while iterating over it

src/qubx/connectors/ccxt/exchange_manager.py
bogdanKaftanatiy: right now `get_ccxt_exchange` returns exchange manager, and exchange manager is the only place where real ccxt exchange is created
and when we recreate exchange, we don't need to create new exchange manager, so either `get_ccxt_exchange` should not return exchange manager, and then another factory method needed for exchange manager, or all exchange creation logic needs to be stored inside exchange manager
I can move some of these methods to a separate factory class if it makes it easier to read

src/qubx/connectors/ccxt/factory.py
```diff
@@ -9,6 +9,7 @@
 
 from .account import CcxtAccountProcessor
 from .exchanges import CUSTOM_ACCOUNTS, CUSTOM_BROKERS, EXCHANGE_ALIASES
+from .exchange_manager import ExchangeManager
 
 
 def get_ccxt_exchange(
```
yuriy-arabskyy, line 15: maybe it makes sense to have something like a separate get_ccxt_exchange and get_ccxt_exchange_manager methods. So that we still have the isolated ccxt exchange creation patter and the manager can use that

src/qubx/connectors/ccxt/handlers/funding_rate.py
```diff
@@ -51,31 +51,34 @@ async def watch_funding():
...
-
-                            # Always emit funding rate
-                            channel.send((instrument, DataType.FUNDING_RATE, funding_rate, False))
-
-                            # Emit payment if funding interval changed
-                            if self._should_emit_payment(instrument, funding_rate):
-                                payment = self._create_funding_payment(instrument)
-                                channel.send((instrument, DataType.FUNDING_PAYMENT, payment, False))
+                            # Handle subscription-specific emissions
+                            if sub_type == "funding_rate":
```
yuriy-arabskyy: revert back to emit both

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,342 @@
...
+# Custom stall detection thresholds (in seconds)
+STALL_THRESHOLDS = {
+    'funding_payment': 12 * SECONDS_PER_HOUR,    # 12 hours = 43,200s
+    'open_interest': 30 * 60,                    # 30 minutes = 1,800s
+    'orderbook': 5 * 60,                         # 5 minutes = 300s
+    'trade': 15 * 60,                            # 15 minutes = 900s
+    'liquidation': 12 * SECONDS_PER_HOUR,        # 12 hours = 43,200s
+}
+DEFAULT_STALL_THRESHOLD_SECONDS = 2 * SECONDS_PER_HOUR  # 2 hours = 7,200s
+OHLC_THRESHOLD_MULTIPLIER = 3  # OHLC threshold = timeframe × 3
```
yuriy-arabskyy: set to 15min for ohlc

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,342 @@
...
+DEFAULT_RESET_INTERVAL_HOURS = 6.0
+SECONDS_PER_HOUR = 3600
+
+# Custom stall detection thresholds (in seconds)
+STALL_THRESHOLDS = {
+    'funding_payment': 12 * SECONDS_PER_HOUR,    # 12 hours = 43,200s
+    'open_interest': 30 * 60,                    # 30 minutes = 1,800s
+    'orderbook': 5 * 60,                         # 5 minutes = 300s
+    'trade': 15 * 60,                            # 15 minutes = 900s
+    'liquidation': 12 * SECONDS_PER_HOUR,        # 12 hours = 43,200s
```
yuriy-arabskyy: set to 7 days

src/qubx/connectors/ccxt/exchange_manager.py
```diff
@@ -0,0 +1,342 @@
...
+DEFAULT_MAX_RECREATIONS = 5
+DEFAULT_RESET_INTERVAL_HOURS = 6.0
+SECONDS_PER_HOUR = 3600
+
+# Custom stall detection thresholds (in seconds)
+STALL_THRESHOLDS = {
+    'funding_payment': 12 * SECONDS_PER_HOUR,    # 12 hours = 43,200s
+    'open_interest': 30 * 60,                    # 30 minutes = 1,800s
+    'orderbook': 5 * 60,                         # 5 minutes = 300s
+    'trade': 15 * 60,                            # 15 minutes = 900s
```
yuriy-arabskyy: set to 1h

PR #89, opened 2025-09-02
Bohdan/statarb

bogdanKaftanatiy:


src/qubx/connectors/ccxt/subscription_manager.py
```diff
@@ -151,6 +151,21 @@ def get_subscription_stream(self, subscription_type: str) -> str | None:
...
+        """
+        Extract the base subscription type without parameters.
+        
+        Args:
+            subscription_type: Full subscription type (e.g., "orderbook(0.0, 20)")
+            
+        Returns:
+            Base subscription type (e.g., "orderbook")
+        """
+        paren_index = subscription_type.find('(')
```
yuriy-arabskyy: why not just use the methods from DataType: `DataType.from_str(subscription_type)[0]` , returns the base type

src/qubx/core/mixins/universe.py
```diff
@@ -50,6 +51,9 @@ def __init__(
         self._instruments = set()
         self._removal_queue = {}
 
+        # - special position gatherer for warmup period
+        self._warmup_position_gathering = SimplePositionGatherer()
```
yuriy-arabskyy: I don't like this duplication in both processing and universe. Would be better to create this in the StrategyContext and maybe put it into the strategy state or some place accessible from both universe and processing, such that we can more easily update the code later without duplication

yuriy-arabskyy (approved):
Once you fix the current comments, I think it's ok to merge

PR #90, opened 2025-09-11
statarb fixes

bogdanKaftanatiy:
Separated sync and async cancel order operations.
Added update order operation. 
Added auto reduce only flag for reduce position orders


src/qubx/backtester/broker.py
```diff
@@ -64,20 +64,66 @@ def send_order_async(
...
+            raise OrderNotFound(f"Order {order_id} not found")
+
+        # Validate that it's a limit order
+        if existing_order.type.lower() != "limit":
+            raise BadRequest(
+                f"Order {order_id} is not a limit order (type: {existing_order.type}). Only limit orders can be updated."
+            )
+
+        # Cancel the existing order first
+        self.cancel_order(order_id)
```
yuriy-arabskyy, line 113: so you decided to always cancel and not use the amend functionality?

yuriy-arabskyy: ah sorry I didn't notice it's simulation

src/qubx/connectors/ccxt/broker.py
```diff
@@ -298,8 +312,19 @@ def _prepare_order_payload(
             logger.warning(f"[<y>{instrument.symbol}</y>] :: Quote is not available for order creation.")
             raise BadRequest(f"Quote is not available for order creation for {instrument.symbol}")
 
-        # TODO: think about automatically setting reduce only when needed
-        if not (reduce_only := options.get("reduceOnly", False)):
+        # Auto-detect if order reduces existing position
+        reduce_only = options.get("reduceOnly", False)
+        if not reduce_only:
```
yuriy-arabskyy, line 317: this is problematic because we could have an opposide side opening order, I think you need to check if the amount by which we enter only decreases exposure and does not open from the other side

yuriy-arabskyy (approved):
looks good with some minor comments, only opposide side entry is important to handle properly

src/qubx/connectors/ccxt/broker.py
yuriy-arabskyy: fixed it

PR #110, opened 2025-12-18
fixed remaining instruments resubscription

bogdanKaftanatiy:


src/qubx/connectors/ccxt/data.py
```diff
@@ -170,17 +170,8 @@ def unsubscribe(self, subscription_type: str, instruments: list[Instrument]) ->
...
-                self._subscription_orchestrator.execute_subscription(
-                    subscription_type=subscription_type,
-                    instruments=remaining_instruments,
-                    handler=handler,
-                    exchange=self._exchange_manager.exchange,
-                    channel=self.channel,
-                    **_params,
-                )
+            # important to update subscription manager
+            self.subscribe(subscription_type, list(remaining_instruments), reset=True)
```
bogdanKaftanatiy, line 174: @yuriy-arabskyy I think the main issue was with manual subscription here, but it is important to update subscription manager, so on the next subscription, when we do union of new instuments and current instruments it should be correct

yuriy-arabskyy:
it looks good, just check the failing test

PR #113, opened 2026-01-09
Implement persistence API

yuriy-arabskyy:
> [!NOTE]
> Introduces a persistence layer for strategy state and wires it through context and runner.
> 
> - Adds `IStatePersistence` protocol and exposes `ctx.persistence`; `StrategyContext` accepts `state_persistence` (defaults to `DummyStatePersistence`)
> - New `qubx.state` module with `DummyStatePersistence` and `RedisStatePersistence` (namespaced keys, optional TTL)
> - Runner updates: config gains `live.state_persistence`, factory creates persistence via `create_state_persistence`, and passes it to `StrategyContext`
> - Config loader now recursively resolves env vars and supports `env:{VAR}` and `env:{VAR:default}` formats
> - Tests: unit tests for dummy, integration tests for Redis (with docker-compose), and updated exporter/env-var tests
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a68cf98adc49aedee9f9a2410aef73f3bfe13c07. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/qubx/core/context.py
```diff
@@ -148,6 +149,7 @@ def __init__(
         health_monitor: IHealthMonitor | None = None,
         restored_state: RestoredState | None = None,
         data_throttler: "InstrumentThrottler | None" = None,
```
dmarienko, line 152: Not sure why we need such quoted syntax ? And also why this TYPE_CHECKING in imports ?

dmarienko (commented):
So how we need to handle state store/restore - manually in code ? 
What if we add another level under this state persistence:
1. add methods store() -> dict[str, Any] / restore(state: dict[str, Any]) to IStrategy, so in strategy we just define what we want to store / restore
2. do state store once per some time (may be add another schedule for it)
3. on start call strategy.restore(xxx) with persisted data
4. May be we should add method for reset state - let's say if we want to run strategy from scratch etc ?

src/qubx/core/context.py
yuriy-arabskyy: if I don't do this, I think there will be a circular import, so TYPE_CHECKING is just to avoid this. I could try to refactor it ofc

yuriy-arabskyy:
> So how we need to handle state store/restore - manually in code ? What if we add another level under this state persistence:
> 1. add methods store() -> dict[str, Any] / restore(state: dict[str, Any]) to IStrategy, so in strategy we just define what we want to store / restore
> 2. do state store once per some time (may be add another schedule for it)
> 3. on start call strategy.restore(xxx) with persisted data
> 4. May be we should add method for reset state - let's say if we want to run strategy from scratch etc ?

I definitely agree with a reset state method to be able to start from scratch. We could also add a store/restore, but this is really quite rare that we need this and it's ok to have this fine-grained control over the store/restore via ctx

dmarienko:
I have a few concerns about fine grained control:
1. you must keep in mind when you want to store / restore - may be OK, but adds additional lines / logic
2. if we put storing state in some high frequency part of code (by mistake or intentionally) it may lead to additional latencies

So having store / restore on strategy level don't make us to think about when / what store or restore.

yuriy-arabskyy:
> I have a few concerns about fine grained control:
> 
> 1. you must keep in mind when you want to store / restore - may be OK, but adds additional lines / logic
> 2. if we put storing state in some high frequency part of code (by mistake or intentionally) it may lead to additional latencies
> 
> So having store / restore on strategy level don't make us to think about when / what store or restore.

I understand that, but what if I want to store state with some irregular frequency when something happens and it's not ok to wait until the end of the specified interval

dmarienko:
Ah, for sure we would have existing 'low level' store / restore mechanism.
What I mean by strategy level - we just create what we want to store and add logic what/how we want to restore.
```
def store(self, ctx):
    return {
        "last-nimble-state": self.xxxx,
        "last-read-time": ....
    }

def restore(self, ctx, state: dict):
   if state["last-read-time"] < ctx.time() + "1h":
      # do something
```

Issue #117, opened 2026-01-14
Switch project to uv


dmarienko:
Let's try uv as modern/flexible and much faster alternative.

PR #122, opened 2026-01-22
Add uv support

yuriy-arabskyy:


yuriy-arabskyy:
Closes #123 and closes #117

PR #193, opened 2026-03-06
feat(runner): pass account_manager to storage constructors

bogdanKaftanatiy:
Adds inspect.signature filtering to construct_storage() so runtime objects like account_manager reach storages that need them (e.g. LighterStorage) while being safely skipped for others. Also adds metadata parameter to export_position_changes for source attribution.


src/qubx/exporters/composite.py
```diff
@@ -5,7 +5,7 @@
 to multiple exporters.
 """
 
-from typing import List
+from typing import Any, List
```
dmarienko: let's get rid of using old style typing List/Dict and switch to modern list/dict instead.

yuriy-arabskyy (approved):

dmarienko (approved):

PR #218, opened 2026-03-25
fix(questdb): add futures exchange table mappings and handle Arrow schema mismatches

bogdanKaftanatiy:


src/qubx/data/cache.py
```diff
@@ -270,16 +270,18 @@ def _merge_batches(existing: pa.RecordBatch, incoming: pa.RecordBatch, time_col_
         [
             pa.Table.from_batches([existing]),
             pa.Table.from_batches([incoming]),
-        ]
+        ],
+        promote_options="permissive",
```
yuriy-arabskyy, line 274: @claude what does this do?

yuriy-arabskyy (approved):

src/qubx/data/cache.py
bogdanKaftanatiy: also fixed an error where some data columns are null, like taker buy volume for kraken candles

PR #220, opened 2026-03-26
fix(trading): round order size up when below min notional threshold

bogdanKaftanatiy:


src/qubx/core/mixins/trading.py
```diff
@@ -563,16 +563,25 @@ def _get_min_size(self, instrument: Instrument, amount: float | None = None) ->
         )
 
     def _adjust_size(self, instrument: Instrument, amount: float) -> float:
-        size_adj = instrument.round_size_down(abs(amount))
+        abs_amount = abs(amount)
+        size_adj = instrument.round_size_down(abs_amount)
```
yuriy-arabskyy, line 567: btw the round size down handles also signed amounts, or did you notice some issue with this? @bogdanKaftanatiy

yuriy-arabskyy: ah the size is made absolute in any case

yuriy-arabskyy (approved):

PR #277, opened 2026-05-06
fix(connectors): restore binance UM watch flags clobbered by ccxt 4.5.52

bogdanKaftanatiy:
## Summary

ccxt 4.5.52 ([ccxt#28493](https://github.com/ccxt/ccxt/pull/28493), released 2026-05-05) introduced an upstream regression in `pro/binanceusdm.py` that wipes out `has["watchOHLCV"]`, `has["watchOHLCVForSymbols"]`, and ~16 other `watch*` capability flags. Qubx handlers gate on these flags during subscription setup and raise `NotSupported` when they're falsy — bots stuck in warmup forever.

This PR extends the existing `BinanceQVUSDM.describe()` override to restore the 5 flags Qubx handlers actually check. Methods themselves still work; only the capability advertisement was broken.

## Reproducer

```python
import ccxt.pro as cxp
ex_50 = cxp.binanceusdm({})  # at ccxt 4.5.50/4.5.51
ex_50.has["watchOHLCV"]              # True
ex_50.has["watchOHLCVForSymbols"]    # True

ex_52 = cxp.binanceusdm({})  # at ccxt 4.5.52
ex_52.has["watchOHLCV"]              # None  ← regression
ex_52.has["watchOHLCVForSymbols"]    # None  ← regression
```

## Why upstream

ccxt 4.5.52 changed `pro/binanceusdm.py:describe()` to merge `async_support.binanceusdm.describe()` ON TOP of pro's describe via `deep_extend`. The REST class explicitly sets every `watch*` flag to `None` (REST doesn't watch). When merged on top, those `None`s overwrite pro's `True`s. Same shape in `pro/binancecoinm.py`, but Qubx doesn't currently use `binance.cm` so no fix needed there.

Plain `cxp.binance` (spot) is unaffected.

### Production code (1 file, +21 lines)

`src/qubx/connectors/ccxt/exchanges/binance/exchange.py` — extend the existing `BinanceQVUSDM.describe()` override (which was already setting `watchTrades.name=aggTrade`) to also restore the 5 watch flags Qubx handlers gate on:

- `watchOHLCV` — handlers/ohlc.py:61
- `watchOHLCVForSymbols` — handlers/ohlc.py:60
- `watchOrderBookForSymbols` — handlers/orderbook.py:101
- `watchTradesForSymbols` — handlers/trade.py:76
- `watchBidsAsks` — handlers/quote.py:46

`BinancePortfolioMargin` and `BINANCE_UM_MM` inherit the fix via `super().describe()`.

### Tests

- **`tests/qubx/connectors/ccxt/test_binance_has_restore.py`** (new, +62 lines) — offline regression test asserting all 5 flags are `True` on `BinanceQVUSDM`, `BinancePortfolioMargin`, `BINANCE_UM_MM`. Runs in CI in <2s without binance network access.
- **`tests/e2e/connectors/ccxt/ccxt_integration_test.py`** (+15 lines) — added a docstring to the existing `test_binance_um_reader` documenting it as the e2e regression test (real binance, local-only). No code change.

## Verification

I ran 4 layers locally before opening this PR. Reviewers can rerun any of them.

### Layer 1 — offline unit test (CI-runnable)

```bash
cd ~/devs/Qubx
.venv/bin/pytest -xvs tests/qubx/connectors/ccxt/test_binance_has_restore.py
# 1 passed in ~2s
```

### Layer 2 — fix works across ccxt versions

```bash
for v in 4.5.50 4.5.52; do
    uv pip install --quiet "ccxt==$v"
    .venv/bin/python -c "import ccxt; print(ccxt.__version__)"
    .venv/bin/pytest -q tests/qubx/connectors/ccxt/test_binance_has_restore.py
done
uv pip install --quiet --upgrade ccxt
.venv/bin/pytest -q tests/qubx/connectors/ccxt/test_binance_has_restore.py
# Restore the lock when done:
uv sync --extra connectors
```

All three pass: 4.5.50 (no-op for the patch), 4.5.52 (broken without patch), latest (currently 4.5.52).

### Layer 3 — sanity check: test fails without the fix

```bash
git stash push -- src/qubx/connectors/ccxt/exchanges/binance/exchange.py
.venv/bin/pytest -xvs tests/qubx/connectors/ccxt/test_binance_has_restore.py
# FAILED — 15 flag×class combinations report None instead of True
git stash pop
```

### Layer 4 — e2e strategy-boot smoke test (local, real binance)

This is the verification that proves the original bug scenario is fixed, not just the `has[]` flag.

```bash
# Force the broken ccxt version to be installed (surfaces the bug):
uv pip install --upgrade 'ccxt'
.venv/bin/python -c "import ccxt; print(ccxt.__version__)"

# Boot a paper-mode binance UM strategy via run_strategy(), subscribe to
# OHLC 1m for [BTCUSDT, ETHUSDT], assert data flows. Without the patch
# this fails with NotSupported during subscription setup.
.venv/bin/pytest -xvs \
    tests/e2e/connectors/ccxt/ccxt_integration_test.py::TestCcxtIntegration::test_binance_um_reader
```

Requires real binance API access — Qubx CI is blocked by Binance HTTP 451, so this runs locally only. **I have not run this myself yet** (would appreciate someone with a local working binance connection running it before merge).

## Out of scope (deliberately)

| | Why |
|---|---|
| `BinanceQVCOINM` (binance.cm wrapper) | Latent bug exists upstream but no current Qubx code path uses `binance.cm`. Easy to add later if needed. |
| `uv sync --frozen` in `qubx deploy` | Lockfile-policy decision; separate concern from this bug fix. Worth considering separately — without it, every strategy is exposed to similar ccxt drift on each pod restart. |
| Tighten `qubx[connectors]` to exclude 4.5.52 | The describe() patch makes 4.5.52 work. Pinning would linger in pyproject indefinitely. |
| Restore all ~18 broken watch flags | Only restoring the 5 Qubx handlers actually check. Reduces "claims True but binance dropped support" surface. |
| Refactor handlers to `try/except NotSupported` instead of `has[]` | Better long-term immunity to all future has[] regressions, but bigger blast radius. Park; revisit if a similar regression hits a non-binance exchange. |

## Production impact

This unblocks `binance.UM`-based strategies on prod (currently `binance.factors`, `binance.factors-uae`, `binance.reversals`). Once a new Qubx tag (1.3.6) is released:

1. Drop the `ccxt>=4.5.46,<4.5.52` workaround pin from `xLydianSoftware/xmetals` (added in xmetals#86) and bump `qubx[connectors]>=1.3.6`.
2. xmetals auto-releases v0.6.5; xrelease bumps `prod/binance/reversals.yaml` ref → v0.6.5.
3. Bot redeploy: pip-installs the latest ccxt (no version pin), boots successfully thanks to this Qubx-side fix.

## Test plan

- [x] Layer 1: offline unit test passes locally
- [x] Layer 2: unit test passes at ccxt 4.5.50, 4.5.52, latest
- [x] Layer 3: unit test fails without the source change (proves it catches the bug)
- [ ] Layer 4: existing `test_binance_um_reader` e2e passes against real binance + ccxt 4.5.52+ (requires reviewer with binance API access)
- [ ] After merge + new Qubx release: redeploy `binance.reversals` bot with the ccxt pin removed from xmetals; bot completes warmup

### DIFF
--- a/src/qubx/connectors/ccxt/exchanges/binance/exchange.py
+++ b/src/qubx/connectors/ccxt/exchanges/binance/exchange.py
@@ -366,15 +366,32 @@ class BinanceQVUSDM(cxp.binanceusdm, BinanceQV):
     def describe(self):
         """
         Overriding watchTrades to use aggTrade instead of trade.
+
+        Also restores watch[*] capability flags clobbered by ccxt 4.5.52's
+        REST-on-pro describe merge in pro/binanceusdm.py (see ccxt PR #28493).
+        The merge runs ``deep_extend(pro.describe(), rest.describe())`` and
+        the REST class explicitly sets every ``watch*`` flag to ``None``,
+        which overwrites pro's ``True`` values. The methods themselves still
+        work; only ``has[]`` was broken. Subclasses (BinancePortfolioMargin,
+        BINANCE_UM_MM) inherit this fix via ``super().describe()``.
+        Remove the ``has`` block below once we bump our minimum ccxt past
+        the regression.
         """
         return self.deep_extend(
             super().describe(),
             {
+                "has": {
+                    "watchOHLCV": True,                # handlers/ohlc.py:61
+                    "watchOHLCVForSymbols": True,      # handlers/ohlc.py:60
+                    "watchOrderBookForSymbols": True,  # handlers/orderbook.py:101
+                    "watchTradesForSymbols": True,     # handlers/trade.py:76
+                    "watchBidsAsks": True,             # handlers/quote.py:46
+                },
                 "options": {
                     "watchTrades": {
                         "name": "aggTrade",
-                    }
-                }
+                    },
+                },
             },
         )
 
--- a/tests/e2e/connectors/ccxt/ccxt_integration_test.py
+++ b/tests/e2e/connectors/ccxt/ccxt_integration_test.py
@@ -372,6 +372,21 @@ class TestCcxtDataProvider:
     @pytest.mark.e2e
     @pytest.mark.skip(reason="Skip by default, run manually if needed")
     async def test_binance_um_reader(self):
+        """Smoke test: paper-mode binance UM OHLC subscription end-to-end.
+
+        Acts as the e2e regression test for the BinanceQVUSDM ``has[]``
+        patch (see ``tests/qubx/connectors/ccxt/test_binance_has_restore.py``).
+        Boots a paper-mode strategy via ``run_strategy()`` with the BINANCE.UM
+        connector, subscribes to OHLC 1m for [BTCUSDT, ETHUSDT], and asserts
+        data flows.
+
+        Without the patch, this test fails at ccxt >= 4.5.52 because
+        ``OhlcDataHandler.prepare_subscription`` raises ``NotSupported`` when
+        the binanceusdm ``has[]`` flags are ``None`` (see ccxt PR #28493).
+
+        Requires real binance API access — Qubx CI is blocked by Binance
+        HTTP 451, so this runs locally only.
+        """
         exchange = "BINANCE.UM"
         await self._test_exchange_reading(exchange, ["BTCUSDT", "ETHUSDT"])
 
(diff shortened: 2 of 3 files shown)


yuriy-arabskyy (approved):

ttodua:
it's a pitty that I was not aware of this issue till date.
upstream is being fixed now and the [fix PR](https://github.com/ccxt/ccxt/pull/28886) will be merged in ccxt soon.
as an advice, it's always better to use the main `binance` class (with `defaultType` option) and never use `binanceusdm, binancecoinm` and alike "alias wrappers".

PR #289, opened 2026-05-28
PR 1 — Phase 1: foundation types, events, bounded channel

bogdanKaftanatiy:
First PR of the account-management redesign (`IBroker`/`IAccountProcessor` → `IConnector` + `AccountManager` + `AccountState`). Implements **Phase 1**: new foundation types added **alongside** the old ones — nothing old is deleted, nothing is wired in yet. Old code keeps working unchanged.

## What's in it (5 commits)
- `feat(events): add OrderStatus enum and OrderOrigin` — `OrderStatus(str, Enum)` (10 states + `is_terminal`/`is_pending`), `OrderOrigin` (FRAMEWORK/RECOVERED/EXTERNAL); old Literal kept as `OrderStatusLegacy`.
- `feat(events): reshape Order and Deal dataclasses` — new `Order` fields (`client_order_id`, `venue_order_id`, `origin`, `filled_quantity`, `avg_fill_price`, `accepted_at`, `rejected_reason`, `last_updated_at`, `retry_count`, `pre_pending_status`, `seen_trade_ids`, `reduce_only`, `post_only`); `Deal.id`→`trade_id`; `Balance = AssetBalance` alias.
- `feat(events): add typed ChannelMessage hierarchy` — `ChannelMessage` → `AccountMessage`/`MarketDataMessage` + all concrete events, `AccountSnapshot`, `ReconcileDiff` (`frozen=True, slots=True`).
- `feat(events): bound CtrlChannel with oldest-drop policy` — bounded queue (default 10_000), oldest-drop, `dropped_count` + `dropped_by_type`.
- `fix(events): serialize CtrlChannel overflow path under lock` — code-review fix.

## Coexistence approach (key design point)
Old code reads **and writes** `order.id` / `order.client_id` and reads `deal.id`, and must not be touched this phase. So `Order` keeps `id`/`client_id` as read/**write** property aliases over `venue_order_id`/`client_order_id`, and `Deal` keeps a read-only `id` alias over `trade_id`. Only **construction sites** were migrated to the new kwargs; all attribute-access (incl. the `BasicAccountProcessor` body) is untouched. All aliases are removed in the cleanup PR.

## Review
- **Spec-compliance review:** ✅ all requirements present, coexistence verified, `BasicAccountProcessor` body unchanged, no over-build.
- **Code-quality review:** one Important issue found (overflow path not thread-safe under concurrent producers) — **fixed** in `ac754a47` by serializing the overflow path under the existing `self.lock`.
- Three deliberate, independently-verified deviations from the literal plan code: `kw_only=True` on events (required or the classes won't define), iterating conformance test instead of `@pytest.mark.parametrize` (repo's `pytest-lazy-fixture` pin breaks parametrize collection), and `CtrlChannel` preserving existing caller-relied behaviors (auto-start, `QueueTimeout` on receive, control guard, non-blocking stop sentinel).

## Tests
`1436 passed, 5 skipped` — full suite green. New: `test_order_status.py`, `test_order_dataclass.py`, `test_events.py`, `test_ctrl_channel.py`.

## Stacking
Base = `account-mgmt-redesign` (long-running integration branch). PR 2 will stack on this branch.


bogdanKaftanatiy:
Addressed review feedback (4 new commits on top, no force-push):

- **`bb682a12` rename `AssetBalance` → `Balance`** repo-wide (73 refs / 16 files); dropped the temporary `Balance = AssetBalance` alias. One canonical name, no shim.
- **`3eee4171` removed unused `pytest-lazy-fixture`** from `pyproject.toml` (it had zero real usages and its plugin broke `parametrize` collection on pytest 8); the events conformance test is now a normal `@pytest.mark.parametrize` (21 cases) with the workaround comment gone.
- **`a39d6b35` dropped unused `OrderStatusLegacy`**; made the `Order.time` comment self-contained (no design-doc reference); converted transitional comments to grep-able `TODO(account-mgmt)` markers (no PR-number/design refs); documented `AccountMessage`/`MarketDataMessage` as marker base classes and the `kw_only=True` rationale.
- **`3fa3fc22` trimmed `test_order_dataclass.py`** to behavior-only (alias round-trips + `require_venue_id`), dropping pure field-storage tests.

`AccountMessage`/`MarketDataMessage` are intentional empty marker base classes — they exist for ProcessingManager's two-level `isinstance` dispatch and the typed `AccountManager.apply(event: AccountMessage)` signature (wired in later PRs), not as DTOs.

Suite: `1454 passed, 5 skipped`. Verified by a focused review of the delta.

src/qubx/backtester/ome.py
```diff
@@ -220,16 +221,17 @@ def place_order(
...
+            client_order_id=client_id,
+            venue_order_id=self._generate_order_id(),
+            origin=OrderOrigin.FRAMEWORK,
+            type=order_type,
+            instrument=self.instrument,
+            time=timestamp,
+            quantity=amount,
+            price=price if price is not None else 0,
+            side=order_side,
+            status="NEW",
```
yuriy-arabskyy: we should use order status enum now everywhere instead of raw strings

src/qubx/backtester/ome.py
```diff
@@ -220,16 +221,17 @@ def place_order(
...
-            timestamp,
-            amount,
-            price if price is not None else 0,
-            order_side,
-            "NEW",
-            time_in_force,
-            client_id,
+            client_order_id=client_id,
+            venue_order_id=self._generate_order_id(),
+            origin=OrderOrigin.FRAMEWORK,
```
yuriy-arabskyy: maybe we should have a default origin FRAMEWORK if not set, because only in rare cases it will be not framework

src/qubx/connectors/ccxt/utils.py
```diff
@@ -78,8 +79,17 @@ def ccxt_convert_order_info(instrument: Instrument, raw: dict[str, Any]) -> Orde
 
     tif = raw.get("timeInForce")
 
+    client_order_id = raw["clientOrderId"]
```
dmarienko: Just in case if raw doesn't have "clientOrderId" - probably we need to use get(...) here ?

tests/qubx/core/test_order_dataclass.py
dmarienko, line 1: It's OK, but just curious - why we literally testing Order's dataclass setters/getters here ;) ?

src/qubx/core/account.py
```diff
@@ -338,7 +341,6 @@ def process_order_request(self, request: OrderRequest) -> None:
             side=request.side,
             status="PENDING",
```
dmarienko: status in order is declared as `OrderStatus` now but here string "pending" is used / no "pending" status defined.

PR #290, opened 2026-05-29
PR 2 — Phase 2: IConnector + AccountManagerConfig + IStrategy callbacks

bogdanKaftanatiy:
**Stacked on #289 (PR 1).** Base is `am/01-foundation-types`, so this diff shows only Phase 2's changes. Review #289 first.

Phase 2 of the account-management redesign: defines the new connector contract, AM config, and strategy callback surface. **Additive only** — nothing old deleted, nothing wired in. Old code keeps working.

## What's in it (4 commits)
- `feat(connector): add IConnector protocol` — `src/qubx/core/connector.py`: `@runtime_checkable` `IConnector` (exchange-facing adapter). `submit_order`→`None`; keyword-only dual-optional `cancel_order`/`update_order` (connector prefers `venue_order_id`); void `request_order_status`/`request_snapshot` (results ride the channel); lifecycle (`connect`/`disconnect`/`is_ws_ready`/`force_ws_reconnect_sync`); `make_client_id`; `is_simulated_trading`/`read_only`; `set_instrument_leverage`/`set_margin_mode`.
- `feat(account): add AccountManagerConfig and _ms_to_cron helper` — `src/qubx/core/account_manager_config.py`: 9 ms-keyed knobs (inflight/snapshot/liveness/terminal-retention) + `_ms_to_cron` (the single source of truth for the 6-field seconds-at-end cron convention).
- `feat(strategy): add typed order/account callbacks to IStrategy` — 13 new ctx-first callbacks (`on_order_accepted/rejected/partially_filled/filled/canceled/expired/updated`, `on_order_cancel_rejected`/`on_order_update_rejected` defaulting to logged warnings, `on_position_update`/`on_balance_update`/`on_funding_payment`, `on_reconcile_complete`). Old `on_order_update`/`on_deals` retained until the cleanup PR.
- `fix(account): reject non-positive intervals in _ms_to_cron` — code-review fix.

## Design notes
- `IConnector`'s **sync-raise vs async-emit** rule is documented in the plan's rejection-boundary table: framework-side validation raises; venue verdicts ride the channel as `OrderRejectedEvent` even when the venue returns a sync REST error.
- Coexistence: `IStrategy` is a regular class (`metaclass=Mixable`), not a Protocol, so adding default-bodied callbacks is safe inheritance — verified no impact on existing strategies.

## Review
- **Spec-compliance:** ✅ all methods/knobs/callbacks present, ctx-first, old callbacks retained, no over-build.
- **Code-quality:** one Important issue (`_ms_to_cron(0)`/negatives emitted invalid cron strings) — **fixed** in `2c728790` with a `<= 0` guard + test, plus a clearer boundary error message.

## Tests
`1465 passed, 5 skipped` — full suite green. New: `test_connector_protocol.py`, `test_account_manager_config.py`, `test_istrategy_callbacks.py`.


src/qubx/core/connector.py
```diff
@@ -0,0 +1,44 @@
...
+from qubx.core.basics import CtrlChannel, Instrument, OrderRequest
+from qubx.core.events import ChannelMessage
+
+
+@runtime_checkable
+class IConnector(Protocol):
+    channel: CtrlChannel
+    exchange_name: str
+
+    def send(self, event: ChannelMessage) -> None: ...
```
yuriy-arabskyy: send is not part of the protocol, it's just a helper method, could be in a separate class and concrete connectors can inherit to get it

src/qubx/core/connector.py
```diff
@@ -0,0 +1,44 @@
...
+                     price: float | None = None,
+                     quantity: float | None = None) -> None: ...
+
+    def request_order_status(self, *, client_order_id: str | None = None,
+                              venue_order_id: str | None = None) -> None: ...
+
+    def request_snapshot(self) -> None: ...
+
+    def is_ws_ready(self) -> bool: ...
+    def force_ws_reconnect_sync(self) -> bool: ...
```
yuriy-arabskyy: maybe let's just rename `force_ws_reconnect_sync` to `reconnect`

src/qubx/core/connector.py
```diff
@@ -0,0 +1,44 @@
...
+    def connect(self) -> None: ...
+    def disconnect(self) -> None: ...
```
yuriy-arabskyy, line 50: do we really need `disconnect` in the interface, not sure who would use it

src/qubx/core/account_manager_config.py
```diff
@@ -0,0 +1,40 @@
...
+    snapshot_check_threshold_ms: int = 5_000
+
+    liveness_check_interval_ms: int = 5_000
+    liveness_check_threshold_ms: int = 30_000
+
+    terminal_order_retention_ms: int = 30_000
+    terminal_order_history_size: int = 10_000
+
+
+def _ms_to_cron(interval_ms: int) -> str:
```
dmarienko: Why we can't use timedelta_to_crontab from qubx.utils.time instead ?

src/qubx/core/interfaces.py
```diff
@@ -2820,6 +2821,40 @@ def on_deals(self, ctx: IStrategyContext, instrument: Instrument, deals: list[De
...
+            f"[{order.client_order_id}] cancel rejected by venue: {reason}; "
+            f"order is STILL ALIVE at the venue"
+        )
+
+    def on_order_update_rejected(self, ctx: IStrategyContext, order: Order, reason: str) -> None:
+        logger.warning(
+            f"[{order.client_order_id}] update rejected by venue: {reason}; "
+            f"order is STILL ALIVE with prior parameters"
+        )
+
```
dmarienko, line 2734: two things:

1. wouldn't it be better to have let's say single entry point for all orders events like `on_order_event(Order, OrderEventData)` - so we could avoid smearing handling code between multiple methods ? As part side - we always invoke single callback in processing code.
 
2. not sure if it's good idea to put logging into interface methods declaration. By default it would puke out these warnings but if I override it in strategy and don't call super it git silent.

tests/qubx/core/test_account_manager_config.py
```diff
@@ -0,0 +1,35 @@
...
+
+from qubx.core.account_manager_config import AccountManagerConfig, _ms_to_cron
+
+
+def test_default_config():
+    cfg = AccountManagerConfig()
+    assert cfg.inflight_check_interval_ms == 2_000
+    assert cfg.snapshot_check_interval_ms == 30_000
+    assert cfg.liveness_check_interval_ms == 5_000
+    assert cfg.terminal_order_retention_ms == 30_000
```
dmarienko: Do we really need to test defaults at all ;) ???

src/qubx/core/interfaces.py
dmarienko: How this connected to new `on_order_updated` now ? Looks like the same ?

PR #291, opened 2026-05-29
PR 3 — Phase 3a: AccountState + state machine + apply() lifecycle

bogdanKaftanatiy:
**Stacked on #290 (PR 2), which is stacked on #289 (PR 1).** Base is `am/02-interfaces`, so this diff shows only Phase 3a's changes. Review #289 → #290 → this.

First **behavioral** PR of the redesign: the central state machine and pure-data account state. **Additive** — new modules + tests only; `AccountManager`/`AccountState` are NOT yet wired into the runner or ProcessingManager (that's PRs 4–7). Nothing old is deleted.

## What's in it (5 commits)
- `feat(account): add exceptions for new account flow` — extends the existing `exceptions.py` with `InvalidOrderTransition`, `OrderAlreadyTerminal`, `ReadOnlyConnector` (reuses the pre-existing `InvalidOrderParameters`/`OrderNotFound`; forward-ref `"OrderStatus"` to avoid the basics→exceptions cycle).
- `feat(account): add AccountState with read API and mutators` — `src/qubx/core/account_state.py`: pure per-exchange data, read API, package-private mutators, index bookkeeping (`_inflight_index`/`_pending_evict_index`/`_venue_id_index`/`_terminal_history`), `seen_trade_ids` fill dedup + weighted avg-fill.
- `feat(account): add AccountManager skeleton with state machine` — `_LEGAL_TRANSITIONS` (verified an exact row-by-row match to the design's transition table), `transition_order` with `pre_pending_status` capture + `InvalidOrderTransition` enforcement, read API.
- `feat(account): implement AM.apply() for order lifecycle` — all order-lifecycle handlers + `_resolve_or_materialize`/`_materialize_external`/`_revert_from_pending`.
- `fix(account): treat lifecycle events on terminal orders as benign no-ops` — code-review fix (see below).

- **Spec-compliance:** ✅ `_LEGAL_TRANSITIONS` programmatically verified an exact match to the design matrix; all handler side-effect/PENDING_/revert rules correct; deviations justified (the plan's `Deal(id=…)` test ctor was stale — `Deal.id` is now a read-only alias, field is `trade_id`).
- **Code-quality:** found 2 Critical issues (same root cause) — **fixed** in `801aecd8`:
  - **C1** lifecycle handlers transitioned via the unchecked mutator, so a late `OrderCanceled` on a `FILLED` order silently flipped it to `CANCELED`.
  - **C2** handlers `KeyError`'d when an event resolved an order evicted to `_terminal_history` (would become a live crash once PR 4 adds the eviction sweep).
  - Fix: a shared terminal-order guard — late events on already-terminal orders are benign no-ops (no transition, no `active_orders[cid]` mutation), generalizing the design's grace-period rule. Added tests for late fill/cancel/update on FILLED and on evicted orders.
  - Also: `_materialize_external` now warns when an external order has no `venue_order_id` (avoids silent `ext:unknown` collisions).

## Tests
`1544 passed, 5 skipped` — full suite green. New: `test_account_state.py`, `test_account_manager_transitions.py` (all 23 legal transitions + illegal-raise + pre_pending), `test_account_manager_apply.py` (handlers, fill dedup, PENDING_* side-effects, clamp+warning, EXTERNAL materialization, terminal no-ops).


bogdanKaftanatiy:
Fixed a correctness bug in `_handle_partial_fill` (commit `aa35eced`), caught in review.

**Bug:** for a `PENDING_UPDATE` order whose fill exceeded the new target, the handler did `order.filled_quantity = order.quantity` — *reducing* `filled_quantity`. But `filled_quantity` mirrors real, irreversible fills and the position (both built from the same `Deal`s), so reducing it desyncs the order from the position: `sum(filled_quantity)` would no longer equal the position. (It also left `avg_fill_price`, computed against the un-clamped total, inconsistent.)

**Fix:** `filled_quantity` is never reduced. A fill that races a pending modify and pushes `filled_quantity` past the new (smaller) target now emits a warning only (no mutation); the venue resolves the race authoritatively — `OrderUpdated` with the accepted quantity, or `OrderUpdateRejected` because it can't shrink an order below what's already filled. Test renamed/rewritten to assert `filled_quantity` stays intact (not clamped) + warning fires.

Suite: `1544 passed, 5 skipped`. The design doc and plan were updated to drop the clamp rule.

src/qubx/core/account_state.py
```diff
@@ -0,0 +1,105 @@
...
+        return self.balances.get(currency)
+
+    def _add_order(self, order: Order) -> None:
+        self.active_orders[order.client_order_id] = order
+        if order.venue_order_id is not None:
+            self._venue_id_index[order.venue_order_id] = order.client_order_id
+        if not order.status.is_terminal():
+            self._inflight_index.add(order.client_order_id)
+
+    def _transition_order(self, cid: str, new_status: OrderStatus, now: np.datetime64) -> Order:
```
yuriy-arabskyy: `_transition_order` is private method here, but who calls it? if it's called from outside, it should be public

src/qubx/core/account_state.py
```diff
@@ -0,0 +1,105 @@
...
+import numpy as np
+
+from qubx.core.basics import Balance, Deal, Instrument, Order, OrderStatus, Position
+
+
+@dataclass
+class AccountState:
+    exchange: str
+    active_orders: dict[str, Order] = field(default_factory=dict)
+    positions: dict[Instrument, Position] = field(default_factory=dict)
```
yuriy-arabskyy: if we expose active_orders, positions, balances via public methods, then these should all be private I think to reduce confusion

src/qubx/core/account_state.py
```diff
@@ -0,0 +1,105 @@
...
+        else:
+            self._inflight_index.add(cid)
+        return order
+
+    def _set_venue_id(self, cid: str, venue_order_id: str) -> None:
+        order = self.active_orders[cid]
+        order.venue_order_id = venue_order_id
+        self._venue_id_index[venue_order_id] = cid
+
+    def _apply_fill(self, cid: str, fill: Deal, now: np.datetime64) -> Order:
```
yuriy-arabskyy: same with apply fill and possibly others, these should all be public methods

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        OrderStatus.PENDING_CANCEL,
+        OrderStatus.FILLED,
+        OrderStatus.CANCELED,
+        OrderStatus.EXPIRED,
+    },
+}
+
+
+class AccountManager:
+    def __init__(self, *, pm, connectors, strategy, time, cfg: AccountManagerConfig | None = None):
```
yuriy-arabskyy: missing typing info

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    def set_context(self, ctx) -> None:
+        """Wire the IStrategyContext after construction.
+
+        AM-fired callbacks (reconcile, inflight-exhaustion) pass this ctx so their
+        signature matches PM-fired callbacks — no None placeholder.
+        """
+        self._ctx = ctx
+
+    def add_order(self, exchange: str, order: Order) -> None:
+        self._states[exchange]._add_order(order)
```
yuriy-arabskyy: order already contains a reference to the instrument which contains exchange, so there is no need to have a separate exchange parameter

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+                return None
+
+    def _get_state_for_event(self, event):
+        if isinstance(event, AccountSnapshotEvent):
+            return self._states.get(event.snapshot.exchange)
+        if event.instrument is not None:
+            return self._states.get(event.instrument.exchange)
+        return None
+
+    def _resolve_or_materialize(self, state, event):
```
yuriy-arabskyy: typing info

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        cid = getattr(event, "client_order_id", None)
```
yuriy-arabskyy: why do we have to use getattr here and not real properties, I think we should be able to know which event it is

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    # Terminal-order guard (shared by the lifecycle handlers below).
+    # A late event for an order that is already terminal — possibly still in
+    # active_orders during the grace window, possibly already evicted to
+    # _terminal_history — must be a benign no-op: NO status change (a terminal
+    # state has no legal outgoing edge) and NO active_orders[cid] mutation (which
+    # would KeyError for an evicted order). This generalizes the OrderAccepted
+    # grace rule (design "OrderFilled before OrderAccepted" / terminal retention)
+    # to every lifecycle event, and is what keeps a late OrderCanceled from
+    # silently flipping a FILLED order to CANCELED.
+    def _is_late_terminal(self, order) -> bool:
```
yuriy-arabskyy: not sure I understand why this is late terminal and not just access the order.status.is_terminal() , also is_terminal could just be a property

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        return order.status.is_terminal()
+
+    def _handle_partial_fill(self, state, event: OrderPartiallyFilledEvent):
+        order = self._resolve_or_materialize(state, event)
+        if self._is_late_terminal(order):
+            logger.debug(f"late partial-fill on terminal {order.client_order_id}; ignoring")
+            return order
+        if event.venue_order_id and order.venue_order_id is None:
```
yuriy-arabskyy: this if doesn't make sense, maybe you wanted to check is not None? but if event.venue_order_id should be enough

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+            state._set_venue_id(order.client_order_id, event.venue_order_id)
```
yuriy-arabskyy: as commented in the state class, these methods should be public (apply_fill, set_venue_id, transition_order, etc)

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        if venue_id is not None and venue_id in state._venue_id_index:
+            return state.active_orders[state._venue_id_index[venue_id]]
+        if cid is not None:
+            for hist in state._terminal_history:
+                if hist.client_order_id == cid:
+                    return hist
+        return self._materialize_external(state, event)
+
+    def _materialize_external(self, state, event):
+        venue_id = getattr(event, "venue_order_id", None)
```
yuriy-arabskyy: how can this happen, isn't it impossible? if we are materializing some order, then for sure at this point which should have the venue_order_id, it should be a parameter of this method and the venue_id is None is redundant

src/qubx/core/account_state.py
```diff
@@ -0,0 +1,105 @@
...
+        order = self.active_orders[cid]
+        if fill.trade_id in order.seen_trade_ids:
+            return order
+        order.seen_trade_ids.add(fill.trade_id)
```
dmarienko: Not sure if keeping executed trades in order object is good idea at all. So Order instance becomes to be connected to history of trades and must track it somehow ?
May be we need some separate storage where we need to maintain trades-id ?

src/qubx/core/account_state.py
```diff
@@ -0,0 +1,105 @@
...
+        order.filled_quantity += abs(fill.amount)
+        if order.avg_fill_price is None:
+            order.avg_fill_price = fill.price
+        else:
+            total = order.filled_quantity
+            prev = total - abs(fill.amount)
+            order.avg_fill_price = (order.avg_fill_price * prev + fill.price * abs(fill.amount)) / total
```
dmarienko: Also it leads to idea that Order must be extended to the full class (not just dataclass) with it's own incapsulated methods for deriving average_fill price etc, because handling all these calculations outside doesn't look OK as for me.

src/qubx/core/account_manager.py
```diff
@@ -0,0 +1,330 @@
...
+            return order
+        if order.venue_order_id != event.venue_order_id:
+            if order.venue_order_id is not None:
+                state._venue_id_index.pop(order.venue_order_id, None)
+            state._set_venue_id(order.client_order_id, event.venue_order_id)
+        if event.new_price is not None:
+            order.price = event.new_price
+        if event.new_quantity is not None:
+            order.quantity = event.new_quantity
+        order.last_updated_at = self._time.now()
```
dmarienko: Probably introducing `Order::set_state(event: OrderUpdateEvent)` would be better solution to incapsulate all logic / checks ?

PR #292, opened 2026-05-29
PR 4 — Phase 3b: snapshot reconcile + ticks + position/leverage/MtM

bogdanKaftanatiy:
**Stacked on #291 (PR 3) → #290 → #289.** Base is `am/03-account-state-sm`, so this diff shows only Phase 3b's changes.

Extends `AccountManager`/`AccountState` with snapshot reconcile, the three scheduled recovery ticks, and position/funding/leverage/MtM. Still **additive** — not yet wired into the runner/ProcessingManager (PRs 5–7). Nothing old deleted.

## What's in it (4 commits)
- `feat(account): snapshot reconcile and terminal eviction` — `_handle_snapshot` (out-of-order ratchet, grace window for missing orders, per-**order** freshness, EXTERNAL/RECOVERED materialization, `ReconcileDiff`, `on_reconcile_complete`), `_sweep_terminal_evictions`.
- `feat(account): periodic ticks for inflight/snapshot/liveness` — three ticks registered via `pm.schedule(_ms_to_cron(...))`; Tier-1 in-flight retry/exhaustion (SUBMITTED→REJECTED, PENDING_*→revert + `on_order_*_rejected`), Tier-2 snapshot, liveness reconnect.
- `feat(account): position math, funding dedup, leverage, MtM` — `_apply_deal_to_position`, `on_market_quote` (MtM), `_handle_funding_payment` (bucket dedup), leverage/margin aggregates, `SimulationAccountManager` subclass.
- `fix(account): drop broken position/balance freshness, guard NaN funding/leverage, fix balance consistency` — review fixes (below).

The plan's pseudocode was adapted to the real `Position`/`FundingPayment`/`Quote` APIs (`unrealized_pnl()` is a method, `notional_value` property, `update_market_price(ts, price, conv)`, `apply_funding_payment(payment, mark)`, `FundingPayment` has no `.amount`).

- **Spec-compliance:** ✅ reconcile/Tier-1/Tier-2/liveness semantics match the design; all 6 API adaptations verified correct in use.
- **Code-quality:** found 1 **Critical** + 4 **Important** + minor — all **fixed** in `f31e98d2`:
  - **C1 (crash):** `getattr(existing, "last_updated_at", np.datetime64(0))` — `np.datetime64(0)` (no unit) raises `ValueError` under numpy 2.x; crashed the steady-state reconcile path (masked only because tests hit the `existing is None` branch).
  - **I1:** position/balance per-record freshness was backed by a monkey-patched `Position.last_updated_at` set on one path only; balance freshness was dead code. → **Dropped per-record freshness for positions/balances** (the whole-snapshot `as_of` ratchet already rejects stale snapshots, and the venue snapshot is authoritative for size/amount). Per-**order** freshness kept (real field). TODO marker to re-add real per-record freshness when WS Position/Balance events are wired.
  - **I2:** funding on an unmarked position → `bal.total += NaN`, permanently poisoning the balance *and* consuming the dedup bucket. → skip without consuming the bucket when mark is NaN.
  - **I3:** `notional_value` NaN for unmarked positions poisoned `get_*leverage` (consumed by emitters/loggers/sizers). → `nan_to_num` guard.
  - **I4:** funding credited `bal.total` but not `bal.free`, breaking the `free == total − locked` invariant. → update both.
  - Added regression tests for all four (snapshot over existing position/balance, funding/leverage on unmarked positions, balance free/total consistency).

## Tests
`1592 passed, 5 skipped` — full suite green. New: `test_account_manager_snapshot.py`, `test_terminal_retention.py`, `test_account_manager_ticks.py`, `test_account_manager_position.py`, `test_account_manager_leverage.py`, `test_account_manager_mtm.py`.


src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
         target = order.pre_pending_status or OrderStatus.ACCEPTED
         order.pre_pending_status = None
         return state._transition_order(order.client_order_id, target, self._time.now())
+
+    def _handle_snapshot(self, state, event: AccountSnapshotEvent):
+        snapshot = event.snapshot
+        if state._last_snapshot_as_of is not None and snapshot.as_of <= state._last_snapshot_as_of:
```
yuriy-arabskyy: let's have a public accessor for the state._last_snapshot_as_of to not use private methods outside

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+            return None
+        state._last_snapshot_as_of = snapshot.as_of
+
+        grace = np.timedelta64(self._cfg.snapshot_check_threshold_ms, "ms")
```
yuriy-arabskyy: I think this order grace period could be calculated once in the constructor of account manager, no need to redo it in each handle snapshot method call

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        diff = ReconcileDiff(exchange=snapshot.exchange, as_of=snapshot.as_of)
+
+        if snapshot.open_orders is not None:
+            snap_by_vid = {o.venue_order_id: o for o in snapshot.open_orders if o.venue_order_id}
+            for cid, cached in list(state.active_orders.items()):
+                if cached.status.is_terminal():
+                    continue
+                vid = cached.venue_order_id
+                if vid is not None and vid in snap_by_vid:
+                    continue
```
yuriy-arabskyy: is it possible that the order is open in both snapshot and state, but the properties of the order differ? in that case we would miss this

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        # by the _last_snapshot_as_of ratchet above — so an accepted snapshot always
+        # overwrites. No per-record freshness here (unlike orders, where it guards a
+        # fresh fill): Position/Balance carry no reliable last-update timestamp yet.
+        # TODO(account-mgmt): once WS PositionUpdate/BalanceUpdate events are wired
+        # (PR 6/7), add per-record freshness backed by a real timestamp so a snapshot
+        # older than a recent WS update can't clobber it.
+        if snapshot.positions is not None:
+            for snap_pos in snapshot.positions:
+                state._set_position(snap_pos.instrument, snap_pos)
+                diff.positions_updated.append(snap_pos)
```
yuriy-arabskyy: seems weird to me that we add to reconcile diff object each position from the state event even if it is the same as the current state

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        seen.add(bucket)
+        amount = pos.apply_funding_payment(payment, mark)  # updates cumulative_funding/pnl
+        bal = state.get_balance(instrument.settle)
+        if bal is not None:
+            bal.total += amount
+            bal.free += amount  # keep free consistent with total (free == total - locked)
+            state._update_balance(instrument.settle, bal)
+        return payment
+
+    def get_total_capital(self, exchange: str | None = None) -> float:
```
yuriy-arabskyy: let's keep the methods organized, and try to keep public methods at the top and private below

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        # truth for size/amount, and stale snapshots are already rejected wholesale
+        # by the _last_snapshot_as_of ratchet above — so an accepted snapshot always
+        # overwrites. No per-record freshness here (unlike orders, where it guards a
+        # fresh fill): Position/Balance carry no reliable last-update timestamp yet.
+        # TODO(account-mgmt): once WS PositionUpdate/BalanceUpdate events are wired
+        # (PR 6/7), add per-record freshness backed by a real timestamp so a snapshot
+        # older than a recent WS update can't clobber it.
+        if snapshot.positions is not None:
+            for snap_pos in snapshot.positions:
+                state._set_position(snap_pos.instrument, snap_pos)
```
yuriy-arabskyy: I think it is possible that other parts of the code get references to the position object and rely on it being updated, but the set position method in the account state replaces the position object completely, making all references in the rest of the code incorrect. I think if a position object exists for some instrument, it should never be replaced, but instead assigned new values

yuriy-arabskyy: @bogdanKaftanatiy this is very critical

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+                diff.positions_updated.append(snap_pos)
+
+        if snapshot.balances is not None:
+            for snap_bal in snapshot.balances:
+                state._update_balance(snap_bal.currency, snap_bal)
```
yuriy-arabskyy: same issue with reassigning balance objects as positions

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+                diff.balances_updated.append(snap_bal)
```
yuriy-arabskyy: why not have some public methods for ReconcileDiff to add balance and positions instead of accessing via diff.positions_updated and diff.balances_updated

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+    def _update_from_snapshot(self, state, existing, snap_order, as_of):
+        existing.status = snap_order.status
+        existing.filled_quantity = snap_order.filled_quantity
+        existing.avg_fill_price = snap_order.avg_fill_price
+        existing.price = snap_order.price
+        existing.quantity = snap_order.quantity
+        existing.last_updated_at = as_of
+
+    def _sweep_terminal_evictions(self) -> None:
+        grace = np.timedelta64(self._cfg.terminal_order_retention_ms, "ms")
```
yuriy-arabskyy: same here, why not do it in constructor once and use it

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        now = self._time.now()
+        for state in self._states.values():
+            if state._terminal_history.maxlen != self._cfg.terminal_order_history_size:
```
yuriy-arabskyy: accessing private account state from outside, and why is the history size not passed during object construction, why is this needed here

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+        threshold = np.timedelta64(self._cfg.inflight_check_threshold_ms, "ms")
+        for exchange, state in self._states.items():
+            for cid in list(state._inflight_index):
+                order = state.active_orders.get(cid)
+                if order is None or order.status not in (
+                    OrderStatus.SUBMITTED,
+                    OrderStatus.PENDING_CANCEL,
+                    OrderStatus.PENDING_UPDATE,
+                ):
+                    state._inflight_index.discard(cid)
```
yuriy-arabskyy: private access

src/qubx/core/account_manager.py
```diff
@@ -1,12 +1,15 @@
+from collections import deque
+
 import numpy as np
 
 from qubx import logger
-from qubx.core.account_manager_config import AccountManagerConfig
+from qubx.core.account_manager_config import AccountManagerConfig, _ms_to_cron
```
dmarienko: same here - _ms_to_cron can be replaced by utility function

src/qubx/core/account_manager.py
dmarienko: What if we could use handlers dict here instead of switch/case ? Something like that:
```
 self._handlers = {
        OrderPartiallyFilledEvent: self._handle_partial_fill,
        OrderFilledEvent:          self._handle_fill,
        OrderAcceptedEvent:        self._handle_accepted,
        OrderCanceledEvent:        self._handle_canceled,
        OrderExpiredEvent:         self._handle_expired,
        OrderUpdatedEvent:         self._handle_updated,
        OrderRejectedEvent:        self._handle_rejected,
        OrderCancelRejectedEvent:  self._handle_cancel_rejected,
        OrderUpdateRejectedEvent:  self._handle_update_rejected,
        AccountSnapshotEvent:      self._handle_snapshot,
        FundingPaymentEvent:       self._handle_funding_payment,
    }

def handle(self, state, event):
    if (handler:=self._handlers.get(type(event))) is None:
        logger.warning(f"unhandled AccountMessage: {type(event)}")
        return None
    return handler(state, event)
```
So we could see all handlers in one place (in init) and this should be faster compared to match on types ?

src/qubx/core/account_manager.py
dmarienko: Also we have a utility method like to_timedelta or something like that

src/qubx/core/account_manager.py
dmarienko: Would be OK to add types declarations here, and also it'd be cool if we add all declarations outside the init, like
```
class AccountManager:
    _pm: <type>
    _connectors: <????>
    _strategy: IStrategy
    .....  
```

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+    def _base_currency_for(self, state) -> str:
+        if not state.balances:
+            return "USDT"
+        return max(state.balances.values(), key=lambda b: b.total).currency
+
+    def _notional(self, pos) -> float:
+        # notional_value is NaN for an unmarked position; treat as 0 so a single
+        # unmarked position can't poison aggregate leverage (consumed by emitters,
+        # loggers, sizers). A real position gets a mark on its first quote.
+        return float(np.nan_to_num(pos.notional_value))
```
dmarienko: Isn't this too heavyweight conversion just for scalar value from pos ?
I suspect that calling np here is 100x times slower compared to simpler logic for NaN.

src/qubx/core/account_manager.py
```diff
@@ -328,3 +349,307 @@ def _revert_from_pending(self, state, order):
...
+            return 0.0
+        pos = state.get_position(instrument)
+        if pos is None:
+            return 0.0
+        total = self._total_capital_for(state)
+        return abs(self._notional(pos)) / total if total > 0 else 0.0
+
+    def get_net_leverage(self, exchange: str | None = None) -> float:
+        states = [self._states[exchange]] if exchange else list(self._states.values())
+        net = sum(self._notional(p) for s in states for p in s.positions.values())
```
dmarienko: I suspect most of the time we will have single exchange there, so probably summation on single exchange list is overkill.

PR #293, opened 2026-05-29
PR 5 — Phase 4: SimulatedConnector (additive)

bogdanKaftanatiy:
**Stacked on #292 → #291 → #290 → #289.** Base is `am/04-reconcile-ticks`, so this diff shows only Phase 4's changes.

## Re-scoped during execution (please note)
The plan originally bundled "SimulatedConnector + full backtester migration" into one PR. But the backtester can't run end-to-end on the new core until `TradingManager.trade` calls `connector.submit_order`+`am.add_order` and `ProcessingManager` dispatches the typed events — both of which are the **next** PR and are **shared with live trading** (so the switch needs an internal dual-path, not a big-bang). So this PR is **just the additive, isolation-tested `SimulatedConnector`** (like PRs 1–4 added new code before wiring); the runner wiring + main-loop events + deletion of `SimulatedBroker`/`SimulatedAccountProcessor` move into **PR 6 (the backtester switch)**, where they land together with the dual-path TradingManager/ProcessingManager.

## What's in it (3 commits)
- `feat(backtest): add SimulatedConnector wrapping OME` — `SimulatedConnector(IConnector)` wraps the existing OME (`ISimulatedExchange`); submit/cancel/update route through the OME and translate `SimulatedExecutionReport` → typed events; `request_order_status`/`request_snapshot`/`process_market_data`; lifecycle/identity stubs. 11 isolation tests (real OME-backed, events drained off a real `CtrlChannel`). OME unchanged.
- `fix(account): use ITimeProvider.time() instead of the non-existent now()` — **cross-cutting correctness fix surfaced while building the connector:** the new `AccountManager` (21 sites) and `SimulatedConnector` (2) called `self._time.now()`, but the real `ITimeProvider` defines `time()`. This passed only because test stubs defined `.now()`; it would `AttributeError` the moment PR 6 wires a real `SimulatedTimeProvider`. Conformed the new code + test stubs to `.time()`.
- `refactor(backtest): clarify SimulatedConnector dead-branch comments and test hygiene` — review polish: annotate the (currently-unreachable, atomic-OME) `OrderPartiallyFilledEvent` branch and the raw-uppercase OME status literals as forward-compat; type `process_market_data`'s `data`; tighten test drain to `QueueTimeout`.

## OME API adaptations (vs the plan's pseudocode), all verified
`get_open_orders()` is a flat `dict[str, Order]`; `cancel_order` raises `OrderNotFound` (not None) → wrapped; `SimulatedExecutionReport` fields/status strings confirmed; OME fills atomically (no partial-fill path — branch kept as scaffolding); `process_market_data` is a generator yielding only exec'd reports.

- **Spec-compliance:** ✅ full `IConnector` surface, all 6 adaptations correct, event mapping correct (single `OrderUpdatedEvent` with stable cid, no stray cancel), time fix complete, no scope creep.
- **Code-quality:** "Ready, nothing blocking" — implementation is more correct than the plan's sketch; minor clarity items applied in commit 3.

## For PR 6 (the switch) to note
- The `OrderPartiallyFilledEvent` path has zero coverage (OME fills atomically) — exercised once a partial-filling connector exists.
- Reject events put a venue id in the `client_order_id` slot when only `venue_order_id` is known — benign (missing-order path), but PR 6's AM consumer should be aware.
- The `loop` arg is accepted for `IConnector` parity but unused in sim.

## Tests
`1603 passed, 5 skipped` — full suite green.


src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+        time_provider: ITimeProvider,
+        # Accepted for IConnector construction parity with live connectors; simulation has no event loop.
+        loop: asyncio.AbstractEventLoop | None = None,
+    ):
+        self.channel = channel
+        self.exchange_name = exchange.exchange_id
+        self._exchange = exchange
+        self._time = time_provider
+
+    def send(self, event: ChannelMessage) -> None:
```
yuriy-arabskyy: `send` doesn't make sense as a public method

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+        for report in self._exchange.process_market_data(instrument, data):
+            self._emit_from_report(report)
+
+    def _emit_from_report(self, report: SimulatedExecutionReport) -> None:
+        order = report.order
+        status = order.status
+        # Raw UPPERCASE status strings from the OME (not the lowercase OrderStatus enum used elsewhere
+        # in core). The OME emits "OPEN", "CLOSED", and "CANCELED" in practice; "NEW" and "FILLED"
+        # are accepted defensively for forward-compatibility with other exchange adapters.
+        if status in ("OPEN", "NEW"):
```
yuriy-arabskyy: shouldn't we use order status enums everywhere now?

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+    exchange_name: str
+
+    def __init__(
+        self,
+        *,
+        channel: CtrlChannel,
+        exchange: ISimulatedExchange,
+        time_provider: ITimeProvider,
+        # Accepted for IConnector construction parity with live connectors; simulation has no event loop.
+        loop: asyncio.AbstractEventLoop | None = None,
```
yuriy-arabskyy: should we just do ** instead

tests/qubx/backtester/test_simulated_connector.py
```diff
@@ -0,0 +1,262 @@
...
+    return events
+
+
+@pytest.fixture
+def setup():
+    instr = lookup.find_symbol("BINANCE.UM", "BTCUSDT")
+    assert instr is not None
+    time = _TimeService()
+    exchange = get_simulated_exchange("BINANCE.UM", time, ZERO_COSTS)
+    channel = CtrlChannel("sim")
```
yuriy-arabskyy: Simulated connector would be using the simulated ctrl channel which uses method callbacks under the hood instead of an actual queue, so the tests here are incorrect

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+    ):
+        self.channel = channel
+        self.exchange_name = exchange.exchange_id
+        self._exchange = exchange
+        self._time = time_provider
+
+    def send(self, event: ChannelMessage) -> None:
+        self.channel.send(event)
+
+    def submit_order(self, request: OrderRequest) -> None:
```
dmarienko, line 47: How we can pass any additional (custom) options here ?
Order can have some custom parameters (and we use it in simulator for some cases when we want to add additional info to OME)

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+            order_side=request.side,
+            order_type=request.order_type,
+            amount=request.quantity,
+            price=request.price,
+            client_id=request.client_id,
+            time_in_force=request.time_in_force,
+        )
+        self._emit_from_report(report)
+
+    def cancel_order(self, *, client_order_id: str | None = None, venue_order_id: str | None = None) -> None:
```
dmarienko: not clear interface for me - it says that I can pass both client_order_id and venue_order_id simultaneously ?

I'd use either `cancel_order(....., venue_or_client_order_id: str | None = None)` declaration or split this method into `cancel_order_by_client_id(...)` + `cancel_order_by_venue_id(....)`

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+            self._emit_from_report(report)
+
+    def update_order(
+        self,
+        *,
+        client_order_id: str | None = None,
+        venue_order_id: str | None = None,
+        price: float | None = None,
+        quantity: float | None = None,
+    ) -> None:
```
dmarienko, line 93: Same thing here - what if I pass different client/venue _ids there ?

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+        oid = venue_order_id or client_order_id
+        try:
+            old = self._exchange.cancel_order(oid)  # type: ignore[arg-type]
+        except OrderNotFound:
+            old = None
```
dmarienko: Do we really need to create old = None here ? It'd be cleaner just to send message and return from method

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,227 @@
...
+                    instrument=None,
+                    client_order_id=client_order_id or oid,  # type: ignore[arg-type]
+                    reason="update_order: order not found",
+                )
+            )
+            return
+        old_order = old.order
+        # cancel+recreate keeps the original client_order_id, so the strategy sees a
+        # single OrderUpdatedEvent rather than a Canceled+Accepted pair.
+        new = self._exchange.place_order(
```
dmarienko, line 112: Same issue here - we do not pass order's options to new one.

src/qubx/core/account_manager.py
```diff
@@ -202,7 +202,7 @@ def _materialize_external(self, state, event):
             origin=OrderOrigin.EXTERNAL,
             type="LIMIT",
             instrument=event.instrument,
-            time=self._time.now(),
+            time=self._time.time(),
```
dmarienko: Semantic question here - Order.time - is it time of order creation or update when state was updated ? 
Probably we need to specify it in name, `time_created` etc ?

PR #301, opened 2026-06-03
feat(account): account-management redesign

bogdanKaftanatiy:
## Overview

Redesigns order submission, lifecycle, and account-state ownership. Replaces the
`IBroker.send_order(req) -> Order | None` contract, the `IBroker` / `IAccountProcessor`
split, and the string-keyed channel dispatch table with:

- a single **`IConnector`** per exchange (exchange-facing only, holds no account state),
- a central **`AccountManager`** that owns all account state, delegating to a pure
  **`reducer`** (one event → one `AccountState` → `ApplyResult`), a **`reconcile`** module
  (snapshot reconciliation + stuck-order sweep logic), and an explicit order
  **state machine**,
- **typed `AccountMessage` events** on the control channel, dispatched by
  `ProcessingManager` (market data stays on the existing `(instrument, dtype, data,
  is_historical)` tuple path — only account events are typed),
- `PENDING_CANCEL` / `PENDING_UPDATE` states, fill dedup by `trade_id`, terminal-order
  retention, two-tier stuck-order reconciliation, WS position/balance application, and
  external-order handling via `Order.origin`.

Hard cut at the major version: no transition feature flag. The backtester migrates in the
same change — `SimulatedBroker` / `SimulatedAccountProcessor` collapse into one
`SimulatedConnector` emitting the same typed events through a synchronous channel.

## Architecture
```
                    ┌────────────────┐  callbacks: on_order(order, change) /
                    │    Strategy    │             on_execution(instrument, deal) /
                    │   (IStrategy)  │             on_position_change(position)
                    └──▲──────────┬──┘  commands:  ctx.trade / cancel / update
         callbacks     │          │     reads:     ctx.account.get_orders /
                    ┌──┴──────────▼──┐              get_position / get_total_capital /
                    │ StrategyContext │  facade     get_available_margin / get_withdrawable_balance
                    └──┬──────────┬──┘
          commands     │          │   reads (via ctx.account → IAccountViewer)
        ┌──────────────┘          └────────────────────┐
        ▼                                               ▼
   ┌──────────────────┐  add_order BEFORE submit  ┌──────────────────────────┐
   │  TradingManager  │  (sync write, rollback    │      AccountManager      │ • routing + read facade
   │  • gen client id │   on framework reject)    │  • apply → reducer        │ • cross-exchange aggregation
   │  • build Order   │ ────────────────────────▶ │  • inflight/snapshot/    │ • owns periodic ticks
   │    (SUBMITTED)   │                            │    liveness ticks        │ • dict[exchange→AccountState]
   └────────┬─────────┘                           └──┬─────────▲──────┬──────┘
            │ submit / cancel / update     apply(event)│         │      │ schedules ticks
            │ (fire-and-forget)                        ▼         │      ▼
   ┌────────┴─────────┐                       ┌──────────────┐  │  ┌──────────────────┐
   │    IConnector    │  self.send(event)     │  reducer.py  │  │  │ ProcessingManager │
   │  (per exchange,  │ ───────┐              │ pure: event  │  │  │ • drain channel   │
   │   holds no state)│        │              │ → ApplyResult│  │  │ • route account / │
   └────────┬─────────┘        │              └──────┬───────┘  │  │   tuple market    │
            │ REST + WS        │                     │ mutates  │  │ • fire callbacks  │
            ▼                  ▼   ┌──────────────┐   ▼          │  │   (error-isolated)│
        Exchange       ──────────▶│  CtrlChannel │┌──────────────┴┐ └─────────▲─────────┘
                   typed account  │  (unbounded) ││ AccountState  │           │
                   event / tuple  └──────────────┘│ + state_machine│──────────┘
                   market data                    │ + reconcile.py │  typed event / tuple
                                                  └────────────────┘
```

## Component interactions

- **Write path (strategy-initiated).** `ctx.trade` → `TradingManager` builds an
  `Order(SUBMITTED)` and writes it to `AccountManager` **before** calling
  `connector.submit_order(...)`, so a synchronous connector's inline accept/fill resolves
  by `client_order_id` instead of spawning a phantom. A framework-side rejection rolls the
  order back; venue rejections (and a synthetic reject on synchronous failure) arrive as
  events. `cancel` / `update` transition to `PENDING_*` and re-raise on synchronous failure
  after routing a synthetic reject so the reducer reverts.
- **Read path (venue-initiated).** Exchange WS → `IConnector` parses → `self.send(event)`
  → `CtrlChannel` → `ProcessingManager`. Market data dispatches straight to the strategy
  (tuple hot path); each typed `AccountMessage` goes through `AccountManager.apply` →
  **pure `reducer`** (state-machine transition + ledger mutation, returning an
  `ApplyResult`) → PM fires the matching strategy callback off the result, error-isolated.
  Empty `ApplyResult` fields are the suppress signal (deduped/late/terminal/unknown events
  fire nothing). **All `AccountState` writes happen on the processing thread** — the
  single-mutator invariant; `on_fit` / `on_warmup_finished` run inline to preserve it.
- **Reconciliation.** `AccountManager` registers its own ticks via the PM scheduler:
  in-flight check (`request_order_status` for stuck orders, with fetch-before-terminalize),
  snapshot backstop (`request_snapshot`, surgical position reconcile that preserves local
  r_pnl/commissions/funding), and connector liveness (`reconnect` on a dead WS).
  Snapshot-derived order corrections fan out as per-position `on_position_change`
  (a consolidated `on_reconcile_complete` is deferred — see design.md).
- **WS account pushes.** Binance UM position/balance pushes are applied conservatively:
  balance pushes apply absolutely with a per-currency ratchet; position pushes never write
  size (sizes are owned by the deal ledger) — a drift triggers a rate-limited snapshot.
  A covered-delta guard keeps deal-driven and push-driven cash from double-booking.
- **Reads.** `ctx.account` exposes the `IAccountViewer` surface (`AccountManager` implements
  it): orders, positions, balances, and the three capital getters
  (`get_total_capital` / `get_available_margin` / `get_withdrawable_balance` —
  venue-sourced in live, derived in sim). No `ctx.state` namespace.
- **Backtester.** `SimulatedConnector` translates each OME execution report into typed
  events over the synchronous `SimulatedCtrlChannel`; `SimulatedAccountManager` runs the
  reducer/reconcile core but registers no live ticks. Same observable strategy behavior.


src/qubx/backtester/runner.py
```diff
@@ -239,14 +242,40 @@ def _process_strategy(self, instrument: Instrument, data_type: str, data: Any, i
...
         return cc.control.is_set()
 
+    def _send_market_data(self, instrument: Instrument, data_type: str, data: Any, is_hist: bool) -> None:
+        # Market data rides (instrument, d_type, data, is_historical) tuples through process_data.
+        # Funding payment is an account event (the AccountManager books it into balances/position
+        # PnL), so a live (non-warmup) funding payment rides the typed channel as a
+        # FundingPaymentEvent — mirroring the live CCXT funding handler. Warmup funding stays on
+        # the cache-only tuple path (matching main, which never booked historical funding).
+        if not is_hist and DataType.from_str(data_type)[0] == DataType.FUNDING_PAYMENT:
+            self.channel.send(FundingPaymentEvent(instrument=instrument, payment=data))
```
yuriy-arabskyy, line 263: will the strategy still see the funding payment as a market data event on which it can react?

src/qubx/backtester/runner.py
```diff
@@ -319,8 +348,8 @@ def _process_event(self, instrument, data_type, event, is_hist, _run, stop_time)
     def _handle_no_data_scenario(self, stop_time):
         """Handle scenario when no data is available but scheduler might have events."""
         # Check if we have pending scheduled events
-        if hasattr(self.scheduler, "_next_nearest_time"):
-            next_scheduled_time = self.scheduler._next_nearest_time
+        if hasattr(self.scheduler, "next_expected_event_time"):
```
yuriy-arabskyy: why is there this hasattr in the first place? are we not sure about the type of scheduler the backtest runner gets?

src/qubx/backtester/runner.py
```diff
@@ -414,29 +442,6 @@ def _basic_initialization(self):
...
-        # - TimeGuardedStorage clamps stop to current sim time (look-ahead guard).
-        # - CachedStorage uses prefetch_period = full sim duration so that the FIRST read
-        #   for any (dtype, symbols) fetches the entire backtest range in ONE DB query — same
-        #   behaviour as the old upfront prefetch but lazy-triggered per dtype actually used.
-        # - Subsequent reads (across all sim ticks) return from cache → zero DB queries.
-        # - NOTE: PrefetchConfig.aux_data_names / args are no longer needed — caching is
-        #   transparent and only warms dtypes the strategy actually accesses.
-        # _inner_aux = self.data_config.aux_storage or self.data_config.data_storage
-        # _pcfg = self.data_config.prefetch_config
-        # if _pcfg is not None and _pcfg.enabled:
```
yuriy-arabskyy, line 428: @dmarienko was this some leftover that was not used?

src/qubx/backtester/simulated_data.py
```diff
@@ -1142,7 +1151,9 @@ def __next__(self) -> tuple[Instrument | None, str, Timestamped | NoDataContinue
                     return None, "", v, False
 
                 instr, pump, subt = self._instruments[k]
-                data_type = pump.producing_data_type
+                # - parameterized label (ohlc -> ohlc(<tf>)) so typed events carry the
+                #   timeframe; non-OHLC types are returned unparameterized as-is.
+                data_type = pump.event_data_type
```
yuriy-arabskyy: @dmarienko does this make sense? this is the code that you worked on recently

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,285 @@
...
+            # channel as OrderRejectedEvent rather than raising. This is the connector
+            # rejection boundary: the same logical failure must take the same path
+            # everywhere, so venue refusals are always async events, not exceptions.
+            self.send(
+                OrderRejectedEvent(instrument=request.instrument, client_order_id=request.client_id, reason=str(e))
+            )
+            return
+        self._emit_from_report(report)
+
+    def cancel_order(self, *, client_order_id: str | None = None, venue_order_id: str | None = None) -> None:
```
dmarienko: I'm not sure that it's OK to have options for two different id's here. What if I pass client_id and venue_id and they are not consistent ?

Same for all order's related methods.

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,285 @@
...
+        # Prefer the venue id (the OME keys orders by it); fall back to the cid before the ack.
+        oid = venue_order_id or client_order_id
+        if not oid:
+            raise ValueError("cancel_order: client_order_id or venue_order_id is required")
+        try:
+            report = self._exchange.cancel_order(oid)
+        except OrderNotFound:
```
dmarienko, line 96: Btw, does it replicate behavior on real exchange - if we try to do cancel on not existing/executed order on ccxt in same way - would it die silently ?

src/qubx/backtester/connector.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def process_market_data(self, instrument: Instrument, data: Timestamped) -> None:
+        # Single entry point that drives the OME (runner._feed_ome in backtest,
+        # _feed_simulated_connector in paper) with whatever a subscription produces. The OME
+        # matches resting orders against Quote/OrderBook/Trade/TradeArray, so those pass
+        # through unchanged — full book depth / trade-array fidelity. Anything else (an OHLC
+        # Bar, a bare price) isn't matchable, so emulate a tradeable quote from it. Note we
+        # can't just always emulate: emulate_quote_from_data has no TradeArray case and
+        # collapses an OrderBook to a single quote.
+        if isinstance(data, (Quote, OrderBook, Trade, TradeArray)):
```
dmarienko, line 223: why we need to check it here ? I thought emulate_quote_from_data should already return None if input data is not recognized ?

src/qubx/backtester/ome.py
```diff
@@ -276,18 +275,20 @@ def _process_order(self, timestamp: dt_64, order: Order) -> SimulatedExecutionRe
                         _exec_price = _desired_fill_price
                     else:
                         raise SimulationError(
-                            f"Special execution price at {_desired_fill_price} for market order {order.id} cannot be filled because market didn't cross this price on last update !"
+                            f"Special execution price at {_desired_fill_price} for market order {order.venue_order_id} cannot be filled because market didn't cross this price on last update !"
                         )
 
             case "LIMIT":
                 _need_update_book = True
+                assert order.price is not None  # LIMIT always carries a price (validated)
```
dmarienko: I think there is better to throw typed exception - like invalid_order or something like that - otherwise we would see some assert exception without any explanation

src/qubx/backtester/ome.py
```diff
@@ -276,18 +275,20 @@ def _process_order(self, timestamp: dt_64, order: Order) -> SimulatedExecutionRe
...
                 if (_buy_side and order.price >= _c_ask) or (not _buy_side and order.price <= _c_bid):
                     _exec_price = _c_ask if _buy_side else _c_bid
 
             case "STOP_MARKET":
                 # - it processes stop orders separately without adding to orderbook (as on real exchanges)
-                order.status = "OPEN"
+                order.status = OrderStatus.ACCEPTED
                 _stp_order = order
+                assert _stp_order.price is not None  # STOP always carries a price (validated)
```
dmarienko: same here for exception

src/qubx/core/basics.py
```diff
@@ -4,8 +4,8 @@
 from enum import StrEnum
 from functools import cache
 from queue import Empty, Queue
-from threading import Event, Lock
-from typing import TYPE_CHECKING, Any, Literal, Optional, TypeAlias, Union
+from threading import Event
+from typing import Any, Optional, TypeAlias, Union
```
dmarienko: I propose to get rid of old style typings here and may be in other places.

src/qubx/backtester/ome.py
dmarienko: Probably here we need to use enumeration now not str ?

src/qubx/backtester/data.py
```diff
@@ -104,9 +104,12 @@ def subscribe(self, subscription_type: str, instruments: set[Instrument], reset:
...
 
                 last_update = h_data[-1]
-                if last_quote := self._account._exchange.emulate_quote_from_data(i, last_update.time, last_update):  # type: ignore
+                if last_quote := self._exchange.emulate_quote_from_data(i, last_update.time, last_update):  # type: ignore
                     self._last_quotes[i] = last_quote
-                    self._account.process_market_data(last_quote.time, i, last_quote)  # type: ignore
+                    # - prime the OME book so resting orders can match from the start;
+                    #   no orders exist yet at subscribe time, so the reports are empty.
+                    for _ in self._exchange.process_market_data(i, last_quote):  # type: ignore
+                        pass
```
dmarienko: I didn't get why we may need to push same quote many times there ?

src/qubx/backtester/simulated_data.py
dmarienko: tbh that's bit misleading, as I see we want to pass timeframe spec within ohlc types if this DataPump is configured to provide ohlc -> ohlc transformation. There are 2 points: 
- why we need timeframe ?
- if we want to do that I'd not add separate _event_type but think about keep it in producing_data_type to have consistent picture.

src/qubx/backtester/simulator.py
```diff
@@ -28,6 +27,29 @@
...
+def _collect_transfers_log(transfer_manager) -> pd.DataFrame | None:
+    """Build the transfers-log DataFrame from a simulation transfer manager.
+
+    The rewritten SimulationTransferManager exposes ``get_transfers() -> dict[tid, record]``
+    (the old ``get_transfers_dataframe()`` is gone). We reproduce the legacy frame shape:
+    one row per transfer, ``timestamp`` as the index, and the remaining record fields as
+    columns. With no transfers we return an empty frame with the legacy schema (matching
+    the original behavior), and None for a manager that doesn't expose the transfers API.
+    """
+    if transfer_manager is None or not hasattr(transfer_manager, "get_transfers"):
```
dmarienko: I don't understand why we ever need to check this hasattr() ??? Isn't this method declared in the interface ?

src/qubx/backtester/simulator.py
```diff
@@ -28,6 +27,29 @@
 )
 
 
+def _collect_transfers_log(transfer_manager) -> pd.DataFrame | None:
```
dmarienko: I think good place for this method would be utils ?

src/qubx/backtester/runner.py
dmarienko: yeah, can be removed for sure

src/qubx/backtester/runner.py
```diff
@@ -217,20 +218,22 @@ def _process_generated_signals(self, instrument: Instrument, data_type: str, dat
...
-            if q := _account._exchange.emulate_quote_from_data(instrument, t, data):
+            if q := _exchange.emulate_quote_from_data(instrument, t, data):
                 _data_provider._last_quotes[instrument] = q
 
         self.time_provider.set_time(t)
-        cc.send((instrument, data_type, data, is_hist))
+        # match resting orders against this tick before the strategy sees it (see _feed_ome)
+        if not is_hist:
+            self._feed_ome(instrument, data)
+        self._send_market_data(instrument, data_type, data, is_hist)
```
dmarienko, line 227: That's what I really don't like here - processing ome as some separate item ! This is internal mechanism must remains behind the simulation exchange and must not be exposed here. 
I'm also not sure about introducing SimulationConnector and instantiating it here as holder of omes.

src/qubx/backtester/utils.py
```diff
@@ -144,7 +148,13 @@ def register(self, callback):
...
     def send(self, data):
-        # - when data is sent, invoke callback
+        # - dispatch synchronously: typed connector events go to process_event,
+        #   legacy market-data tuples go to process_data. Mirrors the live dispatch rule in
+        #   StrategyContext.__process_incoming_data_loop, deliberately WITHOUT its
+        #   health-monitor labels / error notifier / stop-on-True handling: this is the
+        #   backtest hot path and must stay a bare callback invocation.
+        if isinstance(data, ChannelMessage):
+            return self._callback.process_event(data)
         return self._callback.process_data(*data)
```
dmarienko, line 158: Looks not clean but may be it's temporary until we remove legacy data support ?

src/qubx/connectors/ccxt/data.py
```diff
@@ -1,11 +1,11 @@
 import asyncio
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple
```
dmarienko: Let's get rid of old typings here

src/qubx/connectors/tardis/data.py
```diff
@@ -3,13 +3,13 @@
 import threading
 import urllib.parse
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Dict, Tuple
+from typing import Any, Dict, Tuple
```
dmarienko: Let's move to new style

src/qubx/backtester/runner.py
yuriy-arabskyy: I agree regarding internal mechanism remaining behind simulation exchange, but why do you think introducing `SimulationConnector` is not correct? I would call it `SimulatedConnector` to be precise, but it's the correct approach to have simulation implement the connector interface and not have some separate interface from other exchange connector. Or what did you mean? @dmarienko

src/qubx/backtester/utils.py
yuriy-arabskyy: that's what I assumed as well

src/qubx/backtester/runner.py
dmarienko: Well I'm not against the connector abstraction but I'm sure all internal processing must be moved under that then - like ome etc, even more we'd make ome etc to be private members of such connector.

src/qubx/backtester/connector.py
bogdanKaftanatiy: it doesn't make sense to pass both of them, but client_order_id has more priority. 
it just seems easier to use with these methods when we can call the same method but with different params. And for majority of the cases in the codebase we will be using client_order_id. And venue_order_id is basically needed only to work with unknown orders, and to avoid creating bunch of similar methods with separated params

yuriy-arabskyy: I agree with Bogdan on this, because of this thing I wouldn't like to duplicate every method

yuriy-arabskyy: I agree with Bogdan on this, because of this thing I wouldn't like to duplicate every method

PR #321, opened 2026-06-23
Dm/acc mgmt actions

dmarienko:
New diffs extraction mechanism for snapshots vs state (git way)


src/qubx/core/account_manager/reconcile.py
```diff
@@ -1,374 +1,42 @@
-"""Reconciliation logic: snapshot reconcile + the periodic-sweep decision rules.
+"""Order-transition chokepoint + small leaf helpers shared by the reducer and manager.
```
yuriy-arabskyy: maybe this should be something like helpers.py or utils.py, to not confuse with reconciler?

dmarienko: yes, agree, probably I could remove it at all.

src/qubx/core/context.py
```diff
@@ -794,7 +794,7 @@ def update_order(
         order_id: str | None = None,
         client_order_id: str | None = None,
         exchange: str | None = None,
-    ) -> None:
+    ) -> Order:
```
yuriy-arabskyy: you changed here for the update order to return Order, but I checked in the trading manager and it doesn't return anything @dmarienko

yuriy-arabskyy (approved):

src/qubx/core/context.py
dmarienko: yes, reverted

Issue #131, opened 2026-01-29
Remove explicit OHLC timeframe from data configs


dmarienko:
## Description

Remove ability to specify OHLC timeframe in data config keys (e.g., `"ohlc(1h)": storage`). Strategy should specify subscription type in code.

**Part of**: #116 (IStorage/IReader Migration)
**Depends on**: #130 (Simplify DataDecls_t)

## Current Usage

```python
simulate(strategy, {"ohlc(1h)": reader}, ...)  # timeframe in config
```

## New Usage

```python
# In strategy:
def on_init(self, ctx):
    ctx.set_base_subscription(DataType.OHLC["1h"])

# In config:
simulate(strategy, {"ohlc": storage}, ...)  # no timeframe
```

## Tasks

- [ ] Update config parsing to reject timeframe in keys
- [ ] Add helpful error message guiding to new approach
- [ ] Update documentation
- [ ] Update example / existing strategies configs

## Files to modify

- `src/qubx/backtester/utils.py`
- `src/qubx/utils/runner/configs.py`
- Examples and documentation

PR #330, opened 2026-07-08
feat: funding-payment ownership — connectors own account funding, providers own market tuples

bogdanKaftanatiy:
## What

Makes funding-payment ownership structural: **account connectors are the sole emitters of account funding; data providers emit market tuples only; in simulation, only the AccountManager knows funding exists.**

- **Account funding** = `FundingPaymentEvent(instrument, time, amount)` — the venue-settled cash delta, nothing else. Emitted by each venue's connector through whatever mechanism fits the venue; booked once by the reducer (settle-hour dedup); cash coordinated with venue balance pushes as before.
- **Market funding** = `(instrument, DataType.FUNDING_PAYMENT, FundingPayment, hist)` tuples from data providers, universe-scoped, for strategy consumption (EV inputs). Never books positions. `FundingPayment` remains purely a market-data record type.

### Per venue

- **ccxt/Binance**: the connector runs an internal hourly-aligned income poller (started at `connect`, hour boundary +10min, `since = max(connect_time, now − 2h)` so a transiently failed cycle self-heals; survives the connector's own exchange-recreation reconnects by construction). Books venue-exact income amounts — no rate fetching.
- **Hyperliquid** (exchanges repo): books from its `userFundings` WS alone — venue-exact `usdc` amounts live, WS-gap recovery via the venue's snapshot replay + connect-time watermark, duplicates dedup-dropped.
- **Simulation** (paper + backtest): `SimulatedAccountManager.process_market_funding` computes the settlement from market tuples (the framework is the venue) and books through the same reducer path, only for open positions. Live AM no-ops. No booker component, no StrategyContext wiring, no config.

### Review-round history (all reviewer comments adopted)

The PR evolved through review: the original reconciler-driven `request_funding_payments` verb, its scheduling state, the FUNDING_FEE fast path, the funding-rate join, the `source` tag, the event's payment object, and the `funding_sweep_enabled` config were **all removed** in favor of the above (see commits — each round is a net deletion). Position's unread `funding_payments` history is gone; the reducer's computed path and NaN-mark redelivery contract died with the amount-only event.

## Verification

- Unit: **2111 passed / 5 skipped**; golden funding backtest **bit-identical** across every refactor round (`2.484913900565603`, 8-payment trail element-for-element); ruff clean.
- Constructor grep: `FundingPaymentEvent(` exists in exactly `connectors/ccxt/connector.py` (live poller), `account_manager/manager.py` (sim), and the class definition (+ the HPL connector in the exchanges repo).
- **Live testnet, both venues, 28 settlements venue-exact to the sixth decimal**: 13 hourly HPL settlements per pair via WS (including funding-rate sign flips mid-run); the 00:00Z Binance settlement booked by the internal poller and cross-checked against an independently signed `/fapi/v1/income` query; poller lookback re-fetches and **four** live WS reconnect snapshot replays all dedup-dropped (`funding dedup drop` DEBUG trail); final decomposition: booked `fp_pnl` == HPL venue totals + Binance income exactly (BTC −0.263256, ETH +0.050690).

## Cross-repo train (merge order matters)

1. **This PR** → qubx release.
2. exchanges PR #24 (HPL side: WS venue-exact amounts; its REST sweep path deleted with the verb) — hard ordering: needs this PR's protocol (verb removal) first.
3. xrust #131 (xdata `emit_funding_events` purge, v0.8.0).
4. frab: configs drop the flag.

## Follow-up (tracked)

Restart-spanning settlements: persist `last_funding_time` and seed the reducer's dedup table at restore — no connector plumbing needed under this architecture.


src/qubx/connectors/ccxt/connector.py
```diff
@@ -1215,6 +1217,74 @@ async def _hist_deals_async(self, instrument: Instrument, symbol: str, since: dt
...
+                continue
+            records.append((instrument, symbol, int(ts_ms), float(amount)))
+        if not records:
+            return
+        rates = await self._fetch_funding_rates({symbol for _, symbol, _, _ in records}, since_ms)
+        for instrument, symbol, ts_ms, amount in records:
+            payment = FundingPayment(
+                time=ts_ms * 1_000_000,
+                funding_rate=rates.get((symbol, ts_ms // 3_600_000), math.nan),
+                funding_interval_hours=8,  # informational only — not derivable from income records
```
yuriy-arabskyy: this is just inserting fake data, and also if we get real amounts that were booked from/to the account, why do we need to fetch the funding rates additionally? and how can we be sure that we get the correct rate. Is it not better to refactor to use the raw amounts to book correct amounts from the account

yuriy-arabskyy: so the idea is that why do we need the payment object at all if we have the amount

src/qubx/connectors/ccxt/connector.py
```diff
@@ -1215,6 +1217,74 @@ async def _hist_deals_async(self, instrument: Instrument, symbol: str, since: dt
...
+            )
+            self.send(FundingPaymentEvent(instrument=instrument, payment=payment, amount=amount, source="rest"))
+
+    async def _fetch_funding_rates(self, symbols: set[str], since_ms: int) -> dict[tuple[str, int], float]:
+        # best-effort rate join, one attempt per symbol; keyed by (symbol, settlement hour) —
+        # the same hour-bucket convention the AM funding dedup uses
+        rates: dict[tuple[str, int], float] = {}
+        for symbol in sorted(symbols):
+            try:
+                history = await self._em.exchange.fetch_funding_rate_history(symbol, since=since_ms)
```
yuriy-arabskyy: this is a heavy operation on the rate limits, I would rather omit it if possible

src/qubx/core/account_manager/manager.py
```diff
@@ -254,6 +256,10 @@ def _apply_to_state(self, state: AccountState, event: AccountMessage, now: np.da
         result = reducer.apply(state, event, now)
         if isinstance(event, OrderEvent):
             self._execute(state, rec.on_event(state, event, now))
+        # FUNDING_FEE fast path: a settlement just hit the wallet — schedule a debounced
+        # funding sweep so any WS-missed FundingPaymentEvent is recovered promptly.
+        if isinstance(event, BalanceUpdateEvent) and event.reason == "FUNDING_FEE":
+            rec.on_funding_fee(state.exchange, now)
```
yuriy-arabskyy: you pass here now and as I understand it is used in the since query in the ccxt connector, but isn't this unstable? what if now is slightly after the payment timestamp, isn't it just better to schedule funding payment checks 1h+10min and query 20min history?

src/qubx/core/account_manager/reconciler.py
```diff
@@ -350,9 +367,12 @@ class Reconciler:
     _position_confirm_wait: np.timedelta64  # - ConfirmPositionBySnapshot window
     _order_confirm_wait: np.timedelta64  # - AwaitOrderConfirm window
     _order_confirm_max_retries: int
+    _funding_sweep_enabled: bool  # - False disables the funding sweep entirely (kill switch)
     _tasks: dict[Hashable, Task]  # - opaque key -> task, one per key
     _last_snapshot: dict[str, np.datetime64]  # - per-exchange last snapshot time
     _last_full_snapshot: dict[str, np.datetime64]  # - per-exchange last FULL snapshot request time
+    _funding_due: dict[str, np.datetime64]  # - per-exchange armed sweep deadline (absent = not armed)
```
yuriy-arabskyy: why do we need two dictionaries for funding

src/qubx/core/account_manager/reconciler.py
```diff
@@ -50,6 +50,14 @@
 # re-fetched deals are deduped by trade_id and realize-only-guarded anyway.
 HIST_DEALS_LOOKBACK = np.timedelta64(2, "s")
 
+# FUNDING_FEE push fast path: Binance sends one balance push per asset at a settlement
+# boundary — wait out the burst, then sweep once.
+FUNDING_FEE_DEBOUNCE = np.timedelta64(10, "s")
```
yuriy-arabskyy: do we really need this? and even more so this is a bit exchange specific, why not just query once per hour

src/qubx/core/events.py
```diff
@@ -160,7 +160,13 @@ class BalanceUpdateEvent(AccountMessage):
 
 @msg
 class FundingPaymentEvent(AccountMessage):
+    """``amount`` is the venue-exact signed cash delta to the account (settle currency);
+    when set the reducer books it as-is instead of computing qty×mark×rate. ``source``
+    tags the emitter ("ws"/"rest"/"provider"/"sim") for dedup observability."""
+
     payment: FundingPayment
+    amount: float | None = None
+    source: str = ""
```
yuriy-arabskyy: what is the purpose of source here? looks like some debugging artifact rather than really something that belongs in the event

src/qubx/core/account_manager/reconciler.py
```diff
@@ -397,8 +421,57 @@ def on_tick(self, state: AccountState, now: np.datetime64) -> list[Action]:
                 self._last_full_snapshot[state.exchange] = now
             _log.debug(f"[{state.exchange}] reconcile: snapshot due -> RequestSnapshot(include_orders={full_sweep})")
             actions.append(RequestSnapshot(state.exchange, include_orders=full_sweep))
+        if (sweep := self._funding_sweep(state, now)) is not None:
+            actions.append(sweep)
         return actions + self._dispatch(Tick(), state, now)
 
+    def on_funding_fee(self, exchange: str, now: np.datetime64) -> None:
```
yuriy-arabskyy: so maybe we don't need to fast-track/slow-track and can just remove this

src/qubx/core/account_manager/reconciler.py
```diff
@@ -397,8 +420,48 @@ def on_tick(self, state: AccountState, now: np.datetime64) -> list[Action]:
                 self._last_full_snapshot[state.exchange] = now
             _log.debug(f"[{state.exchange}] reconcile: snapshot due -> RequestSnapshot(include_orders={full_sweep})")
             actions.append(RequestSnapshot(state.exchange, include_orders=full_sweep))
+        if (sweep := self._funding_sweep(state, now)) is not None:
+            actions.append(sweep)
```
dmarienko: Didn't get why we need to arm this as separate action on every tick and not part of whole snapshot ?

src/qubx/core/account_manager/reconciler.py
```diff
@@ -397,8 +420,48 @@ def on_tick(self, state: AccountState, now: np.datetime64) -> list[Action]:
...
+    @staticmethod
+    def _funding_anchor(state: AccountState, now: np.datetime64) -> np.datetime64:
+        # - startup watermark: +1ns past the newest restored settlement (EXCLUSIVE — it is already
+        #   inside restored cumulative_funding and the bucket table is empty after restart, so
+        #   re-fetching it would double-book); no restored anchor -> now (restart gap stays open
+        #   until last_funding_time persistence lands).
+        times = []
+        for pos in state.get_positions().values():
+            t = pos.last_funding_time
+            if isinstance(t, (int, np.integer)):  # in-session bookings stamp the raw ns epoch
```
dmarienko: why not to store this time in pos.last_funding_time as np.dt64 already and avoid these magic conversions every time ?

src/qubx/core/context.py
```diff
@@ -169,6 +169,7 @@ def __init__(
         rate_limiting_config: Any | None = None,
         event_loop: asyncio.AbstractEventLoop | None = None,
         read_only: bool = False,
+        funding_booker: IFundingBooker | None = None,
```
dmarienko: Not sure about this design - what if we'd move IFundingBooker totally inside the AcocuntManagement impl and wouldn't expose in StrategyContext at all ? So only AM knows about any funding (or any other payment in case if it's not crypto) ?

src/qubx/connectors/ccxt/connector.py
```diff
@@ -1215,6 +1217,51 @@ async def _hist_deals_async(self, instrument: Instrument, symbol: str, since: dt
...
+            if symbol is None or ts_ms is None or amount is None:
+                logger.debug(f"[{self.exchange_name}] funding record missing fields, skipped: {raw}")
+                continue
+            try:
+                instrument = self._instrument_for_symbol(symbol)
+            except CcxtSymbolNotRecognized:
+                logger.debug(f"[{self.exchange_name}] funding record for unknown symbol {symbol}, skipped")
+                continue
+            payment = FundingPayment(
+                time=int(ts_ms) * 1_000_000,
```
dmarienko: why we need this time in seconds there ?

src/qubx/utils/runner/runner.py
```diff
@@ -669,6 +670,9 @@ def create_strategy_context(
         rate_limiting_config=_rate_limiting_config,
         event_loop=loop,
         read_only=config.live.read_only,
+        # paper books funding off market funding tuples (no venue account to do it);
+        # live books via the connector's typed events, so no booker
+        funding_booker=SimFundingBooker(_am) if paper else None,
```
dmarienko: same q. here - why we need this SimFundBookr here at all ? Could we create it inside SimAccountManager ?

src/qubx/utils/runner/configs.py
```diff
@@ -194,6 +194,8 @@ class AccountManagerConfig(StrictBaseModel):
 
     position_confirm_wait_ms: int = 2_000
 
+    funding_sweep_enabled: bool = True  # False disables the funding sweep (kill switch)
```
dmarienko: I think it's bit redundant option - AccountManager should take care about it in case if it's funding-related account ?

src/qubx/core/account_manager/reducer.py
```diff
@@ -510,32 +510,25 @@ def _handle_update_rejected(state: AccountState, event: OrderUpdateRejectedEvent
...
+    amount = pos.apply_funding_payment(event.time, event.amount)  # updates cumulative_funding/pnl
     # Skip the cash leg when a venue balance push already covers it (the venue debits
-    # the wallet and pushes the new total with reason FUNDING_FEE — booking our computed
-    # amount on top would double-count). cumulative_funding/pnl above always book.
-    # payment.time is a venue-clock ns epoch — same domain as the push as_of.
-    covered = _covered_by_balance_push(state, instrument.settle, np.datetime64(int(payment.time), "ns"))
+    # the wallet and pushes the new total with reason FUNDING_FEE — booking the amount
+    # on top would double-count). cumulative_funding/pnl above always book.
+    # event.time is venue-clock — same domain as the push as_of.
+    covered = _covered_by_balance_push(state, instrument.settle, event.time)
```
yuriy-arabskyy: is there a possibility that it is not covered by balance update? Maybe we can just safely assume that only funding field needs to be updated and balance will be synced anyway always, so no need to update balance from funding either way and we can just omit this logic

src/qubx/connectors/ccxt/connector.py
```diff
@@ -1215,6 +1218,62 @@ async def _hist_deals_async(self, instrument: Instrument, symbol: str, since: dt
...
+    async def _funding_poll_loop(self) -> None:
+        connect_time = self._time.time()
+        while True:
+            now = self._time.time()
+            await asyncio.sleep(float((self._next_funding_poll_at(now) - now) / np.timedelta64(1, "s")))
+            await self._funding_payments_async(self._funding_since(connect_time, self._time.time()))
+
+    @staticmethod
+    def _next_funding_poll_at(now: dt_64) -> dt_64:
+        return now.astype("datetime64[h]") + np.timedelta64(1, "h") + np.timedelta64(10, "m")
```
yuriy-arabskyy: maybe we can be more precise and actually make sure it's 1h+10min, because now can be whenever

yuriy-arabskyy (approved):

PR #332, opened 2026-07-09
fix(runner): seed live warmup with real account capital via pre-warmup snapshots

bogdanKaftanatiy:
## Problem
In live mode the warmup simulation is seeded with `ctx.account.get_total_capital()` — but connectors only connect at `ctx.start()`, **after** warmup, so the read always returns **0**. The warmup sim then runs the whole strategy against an unfunded account: sizing collapses to zero and any capital-dependent math misbehaves. First observed as a platform bot startup crash: frab's exchange rebalancer divides by average balance → `ZeroDivisionError` on every warmup trigger event → `StrategyExceededMaxNumberOfRuntimeFailuresError`. `_validate_warmup_result` then (correctly) rejects the 0-capital outcome, so **no live bot with a `live.warmup` block can currently boot**.

## Root cause — a refactor regression
This used to work: v1.5.0's `CcxtAccountProcessor.get_total_capital()` lazily ran a blocking `fetch_balance` whenever polling hadn't started yet, so warmup always saw real venue capital. The AccountManager refactor made the getter a pure state read (right call — no hidden I/O in getters) but the boot-time balance fetch it silently provided was never re-homed.

Paper was unaffected by ordering luck: `_seed_paper_capital` runs at context construction, before warmup.

## Fix
`_sync_account_snapshot_before_warmup(ctx)`, called right before the warmup sim is seeded:
- requests `request_snapshot(include_orders=False)` from every connector — pure REST, works pre-`connect()` (verified for ccxt and hyperliquid)
- drains the resulting `AccountMessage`s from the databus straight into `AccountManager.apply` until `is_synced()` (bounded, 15s)
- paper skips it; on timeout it proceeds with whatever arrived and `_validate_warmup_result` makes the call

Applying to the live AM pre-warmup is the same state transition the post-start reconciler performs on its first tick — just earlier, and deliberately **without** strategy callbacks (the strategy isn't warmed up yet; boot-time discovery is contractually the `_is_ready` gate + `ctx.get_positions()`, not callbacks). Side benefits: `is_synced()` is already true at live start (readiness gate passes immediately instead of waiting toward `ACCOUNT_SYNC_TIMEOUT`), `ctx.start()` sees venue positions when computing initial instruments, and the post-warmup state resolver compares against real state.

## Verification
- Reproduced the platform crash locally (live testnet + warmup block): 21 ZeroDivisions, warmup abort — then with this fix, same setup: `Warmup capital seeded from account snapshots: 5,798.84` (both venues), all 24 warmup triggers through the **unmodified** frab rebalancer, zero errors, validation passed, live start.
- Degradation path exercised organically: with Binance testnet REST rate-banned (418), the seed proceeded with the reachable venue only (796.32) instead of hanging.
- Unit tests: snapshot pump, paper skip, timeout path, validator (unfunded raises / funded bankruptcy raises). Full runner test file green; ruff clean.

### DIFF
--- a/src/qubx/utils/runner/runner.py
+++ b/src/qubx/utils/runner/runner.py
@@ -966,12 +966,12 @@ def _sync_account_snapshot_before_warmup(ctx: IStrategyContext, timeout: float =
             # pre-start only connectors produce events, so nothing else should flow here
             logger.debug(f"pre-warmup account sync: dropping non-account event {type(event)}")
 
-    if account.is_synced():
-        logger.info(
-            f"<yellow>Warmup capital seeded from account snapshots: {account.get_total_capital():,.2f}</yellow>"
+    if not account.is_synced():
+        raise WarmupValidationError(
+            f"Account snapshots not applied within {timeout:.0f}s — cannot seed warmup capital; "
+            "check venue connectivity and credentials"
         )
-    else:
-        logger.warning("Account snapshots not applied within timeout — warmup may run with incomplete capital")
+    logger.info(f"<yellow>Warmup capital seeded from account snapshots: {account.get_total_capital():,.2f}</yellow>")
 
 
 def _run_warmup(
--- a/src/qubx/utils/runner/runner.py
+++ b/src/qubx/utils/runner/runner.py
@@ -36,7 +36,8 @@ from qubx.core.basics import (
 )
 from qubx.core.connector import IConnector
 from qubx.core.context import StrategyContext
-from qubx.core.exceptions import WarmupValidationError
+from qubx.core.events import AccountMessage
+from qubx.core.exceptions import QueueTimeout, WarmupValidationError
 from qubx.core.helpers import BasicScheduler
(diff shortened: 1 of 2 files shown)


yuriy-arabskyy (approved):

src/qubx/utils/runner/runner.py
```diff
@@ -928,6 +929,51 @@ def _validate_warmup_result(result: WarmupResult, min_coverage_ratio: float = 0.
...
+        else:
+            # pre-start only connectors produce events, so nothing else should flow here
+            logger.debug(f"pre-warmup account sync: dropping non-account event {type(event)}")
+
+    if account.is_synced():
+        logger.info(
+            f"<yellow>Warmup capital seeded from account snapshots: {account.get_total_capital():,.2f}</yellow>"
+        )
+    else:
+        logger.warning("Account snapshots not applied within timeout — warmup may run with incomplete capital")
```
yuriy-arabskyy: should fail here

PR #336, opened 2026-07-15
feat(core): transfer-manager lifecycle fixes — sim pickup, warmup safety, paper pre-seed

bogdanKaftanatiy:
## Summary

Framework-level `ITransferManager` lifecycle fixes so live transfer managers injected by strategies (via `initializer.set_transfer_manager(...)` in `on_init`) are handled safely across simulation, warmup, and paper modes. Per review feedback, the service-specific HTTP client and all config wiring moved out of this (public) repo into our internal package — this PR now contains only generic framework correctness:

- **Simulation pickup fix** (pre-existing bug): the backtester assigned `SimulationTransferManager` *after* `StrategyContext.__post_init__` had already read the initializer — the only pickup point — so plain simulations ran with no transfer manager. The auto-assign now precedes context construction.
- **Warmup safety**: the warmup `SimulationRunner` shares the live context's initializer; an `on_init`-injected live manager would previously have executed **real transfers during warmup**. Now: warmup mode force-assigns a fresh `SimulationTransferManager` (on_init is not re-run in warmup, so nothing overrides it), and `_run_warmup` stashes the live manager and restores it in a widened `try/finally` — a Ctrl-C during warmup construction can no longer leak `simulation=True` or a sim manager into the live context.
- **Paper pre-seed**: `create_strategy_context(paper=True)` seeds `SimulationTransferManager` before construction, so paper twins rehearse `ctx.transfer_funds(...)` against simulated balances by default; a strategy's `on_init` may still override deliberately (its responsibility).
- `set_transfer_manager` widened to accept `None` (needed by the warmup restore).

No new config surface, no new public API beyond the `| None` widening. Second commit shows exactly what moved out.

## Test plan

- New ctx-level regressions: sim context gets a working manager and `transfer_funds` moves balances; warmup force-assigns over a stub live manager; `on_init` injection wins in plain sims; `create_strategy_context(paper=True)` pre-seeds and the context picks it up
- `tests/qubx/core/ + utils/ + backtester/`: 960 passed, 5 skipped; ruff clean on all touched files


src/qubx/backtester/transfers.py
```diff
@@ -29,11 +29,12 @@ def transfer_funds(self, from_exchange: str, to_exchange: str, currency: str, am
...
 
-        # Instant transfer: debit the source and credit the destination. adjust_balance
-        # mutates the held Balance in place (holders keep a live reference) and creates
-        # the destination Balance if missing.
+        # adjust_balance mutates the held Balance in place (holders keep live references) and creates missing ones
         self._account.adjust_balance(from_exchange, currency, -amount)
-        self._account.adjust_balance(to_exchange, currency, amount)
+        # both-stables transfers credit the destination's base currency — mirrors the live service's route asset
+        to_base = self._account.get_base_currency(to_exchange)
+        credit_currency = to_base if {currency, to_base} <= {"USDT", "USDC"} else currency
```
yuriy-arabskyy: was credit currency needed somewhere? I understand where you're coming from, but I don't like it much, because a transfer in one currency should really just be a transfer, and credit currency means that there is some implicit conversion somewhere which makes a bit more complicated than it should be imo

yuriy-arabskyy (approved):

PR #341, opened 2026-07-22
fix(runner): build data providers with the canonical exchange name

yuriy-arabskyy:
## Summary

Live finding on the first BINANCE.PM bot (v11.dynamic.bhpl, dev): the **UniverseManager swept all 673 Binance instruments as "delisted (gone)"** right after warmup.

Root cause: the runner builds the data provider's `BuildContext` with the configured **venue** name. Data plugins key their catalogs and subscription requests by `ctx.exchange_name` — qubx-xdata builds `"EXCHANGE:TYPE:SYMBOL"` keys — so on a `BINANCE.PM` venue the provider asked the xdata catalog for `BINANCE.PM:SWAP:BTCUSDT`, which doesn't exist (the catalog speaks canonical `BINANCE.UM`). `is_instrument_listed` returned False for everything → universe emptied. Live subscriptions through the same provider were equally mis-keyed.

Fix: the DATA provider now gets a canonicalized context (`dataclasses.replace`) while the CONNECTOR keeps the venue context (credentials + papi order routing). Market data is public and belongs to the instrument universe — the same principle the rest of the PM design already follows (#340).

Side effect for `connector: ccxt` on a PM venue: public market data now uses the canonical UM class (`binanceqv_usdm`) instead of `binancepm` — same public endpoints, faster startup (no spot/inverse market loading on the MD instance).

## Test
Extended the R13 canonicalization regression to assert the data-provider context is canonical. Full suite: 2186 passed.

### DIFF
--- a/src/qubx/utils/runner/runner.py
+++ b/src/qubx/utils/runner/runner.py
@@ -5,7 +5,7 @@ import socket
 import tempfile
 import time
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from functools import reduce
 from pathlib import Path
 from threading import Thread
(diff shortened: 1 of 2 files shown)


yuriy-arabskyy:
Declined in favor of fixing in qubx-xdata directly — xdata is the standard data provider for all future deployments, so canonicalizing there covers the practical surface; keeping this branch around in case we later want the invariant at the runner boundary instead (see discussion re: venue_name in BuildContext).

PR #348, opened 2026-07-29
chore: sync dev with main

yuriy-arabskyy:
dev had drifted ~3 months behind main (releases cut from main). Sync so future branches off dev carry current code. Content after merge == main (the determinism fix #347 exists on both sides with identical changes).


yuriy-arabskyy:
Superseded: dev synced to main via -s ours merge commit (fast-forward push); tree now identical to main.

PR #366, opened 2026-08-05
Degraded-state trading + non-blocking leverage

dmarienko:
Two things, both about not letting the framework act on stale or slow information.

## 1. Trading while the context is degraded

`ctx.status` goes DEGRADED when the framework knows it is behind — today that means the event queue has not drained for a while.

While degraded, `trade()` and `update_order()` now raise `QubxDegradedState`. **Every order**, not just the ones that add exposure. `cancel_order()` is never refused: pulling a resting order lowers exposure and needs no fresh data.

The earlier version allowed position-reducing orders. That only makes sense when the reason is our own stale view. When the reason is the venue — maintenance, an outage — a closing order cannot reach it either, and accepting one reports an exit that never happened. A strategy cannot tell the two apart from the flag, and the list of reasons will grow, so refusing everything is the rule that stays correct.

Off by default. Strategies opt in with `initializer.set_deny_trading_when_degraded(True)` (was `set_reduce_only_when_degraded`).

Live: under a flooded event channel the run refused opening orders and accepted closes, before this change made closes refused too.

## 2. Setting leverage no longer blocks the caller

Setting leverage across a wide universe took **127 seconds** of the ProcessorThread, during which no events were processed. Batching cut it to 3.1s. This takes it off the thread entirely — **1.4ms** measured live for 15 instruments.

`set_instrument_leverage` returns nothing and does not wait. The connector sends the venue call on its own loop. A refusal comes back as a `VenueOperationError` and reaches the strategy's `on_error`; it used to be a return value nobody checked.

`set_instrument_leverages` (the batch) is removed. It existed to overlap the waits, and there is nothing left to wait for.

The ccxt connector reads every symbol's configured leverage and venue maximum **once an hour** and keeps them, so:

- a request above the maximum is lowered to it
- a request for the value the venue already has is not sent at all
- whole numbers only — Binance rejects a float leverage as malformed (`-1102`)

Measured live on BINANCE.PM: at leverage 100, of 64 requests **62 were lowered and 2 sent, 0 refused**. At leverage 3, **62 skipped, 2 sent** — the 2 being the symbols genuinely still at 100.

### Renamed getter

`get_max_instrument_leverage` → **`get_instrument_leverage`**, because that is what every implementation returned: the leverage *you configured*, not the highest the venue allows. Its own docstring said "currently configured".

A new `get_max_instrument_leverage` returns the real venue maximum. Both read the hourly cache instead of making a venue call.

`get_instrument_leverage` on the AccountManager now falls back to the connector when the account snapshot has nothing — which is the usual case, since the snapshot only carries leverage for instruments you hold a position in, and on Binance not even then.

**Breaking:** `set_instrument_leverage` returns `None` not `bool`; `set_instrument_leverages` is gone; `get_max_instrument_leverage` means something different. No callers outside this repo used any of them except the setter, which frab updates alongside.

## Also

`VenueOperationError` in `core/errors.py` — for a venue write that fails after its caller has already been released. Rides the existing error path to `IStrategy.on_error`.

## Verification

`tests/qubx`: **1394 passed.**

Live on BINANCE.PM × HYPERLIQUID.F across several runs: the leverage clamp, the skip, the float rejection and the fix for it, and the drop from 15,488ms to 1.4ms on the Hyperliquid side (that connector's half ships in the exchanges repo).

## Needs

Connectors implementing `IConnector` must drop `set_instrument_leverages` and change `set_instrument_leverage` to return `None`. `qubx-hyperliquid` does this in its own branch of the same name.


src/qubx/core/mixins/trading.py
```diff
@@ -72,35 +79,86 @@ def __init__(
...
     def _ensure_writable(self) -> None:
-        """Single read-only gate for the venue-write boundary (trade/cancel/update)."""
+        """Single read-only check for the venue-write boundary (trade/cancel/update)."""
         if self._read_only:
             raise ReadOnlyConnector("trading is read-only — write rejected")
 
+    def set_reduce_only_when_degraded(self, enabled: bool) -> None:
+        self._reduce_only_when_degraded = enabled
+
+    def _reduce_only_amount(self, instrument: Instrument, amount: float) -> tuple[float, float]:
```
yuriy-arabskyy, line 93: why do you need to modify the amount? we could just allow orders with reduce_only flag set and nothing else, that way we don't need to check and modify amounts and we guarantee that only reduce_only goes through which works independent if our internal state is correct or not. Because doing reduce only based on current quantity with a degraded state, could mean that we have incorrect internal state and don't allow passing a valid reduce only order

dmarienko: it's not guaranteed that I won't send reduce only order, what if I send just opposite order to close part / full position ?

yuriy-arabskyy (approved):

PR #367, opened 2026-08-05
feat(account): order-replace orchestration — truthful update_order for partially-filled orders

yuriy-arabskyy:
## What

Makes `ctx.update_order` correct for partially-filled orders on every venue, by fixing the `Order.quantity` bookkeeping invariant and routing quantity updates of partially-filled orders through a framework-side cancel+replace orchestration instead of venue-native amend.

**Contract change:** `update_order(price, amount)` — `amount` is the desired **signed remaining** to keep working (sign must match order side; zero/mismatch raises instead of the old silent `abs()`). `Order.quantity` now always means **total including everything ever filled** on the client order id; the amend-ack reducer splices `quantity = filled_quantity + new_quantity`.

**Routing rule:** `filled == 0` → native venue amend (the one case where every venue dialect agrees). `filled > 0` → framework cancel+replace above the `IConnector` boundary, same client order id: arm a `ReplaceIntent` → internal cancel (suppressed by the reducer; the venue's final fill figure rides the new `OrderCanceledEvent.venue_filled_quantity`, populated by the ccxt connector from the cancel ack) → account-reconciler decision (`SubmitReplacement`/`AbandonReplace` from the post-cancel residual) → `_execute` submits the replacement and routes the splice event.

**Failure contract — truth over fiction, every path:** sync cancel failure → intent cleared + synthetic update-reject (order alive); sync submit failure → truthful CANCELED (order is dead at the venue); residual below min → truthful CANCELED; duplicate REST/WS cancel terminals and stale cancels of superseded venue ids → dropped (per-cid superseded-oid marker); strategy cancel during the replace window wins (no resurrection); unknown-outcome intent expiry (30s) → revert + status refetch, never a fabricated terminal.

## Why

Live incident 2026-08-05 (dev frab bot, pair ACE): repricing a partially-filled Hyperliquid maker order corrupted the Order record — HL's modify is a cancel+replace under the same client id, the amend ack overwrote `quantity` with the remaining while `filled_quantity` stayed cumulative, and `remaining = quantity − filled` went negative. Downstream execution logic then trimmed the wrong leg and livelocked. On Binance/Gate the same call pattern is wrong differently (their amend-quantity means *new total* with executedQty preserved → silent double-shrink). Full analysis: `docs/account-management/design.md` § "Order updates and the replace orchestration" and `docs/superpowers/plans/2026-08-05-order-replace-orchestration.md`.

## Testing

- ~50 new/changed tests: reducer splice + suppression + duplicate/stale-cancel regressions, routing + sign validation, intent lifecycle (arm/clear on every path), decision math (incl. below-min + venue-id-only cancels), `_execute` end-to-end through real reducer+reconciler (incl. sync-failure truth paths, raced-fill lag compensation, ResolveMissingOrder watcher resolution), expiry recovery, and a backtester integration test pinning the `filled == 0` amend path against real OME book state (with an intermediate no-fill quote so a venue-side reprice failure cannot pass).
- Branch suite: `tests/qubx/core` + `tests/qubx/backtester` green (1030 passed). Note: `tests/qubx/connectors/ccxt/test_ohlc_pagination.py` shows 5 **pre-existing** ordering-dependent failures when the connectors suite runs together — verified identical on `main` (event-loop test-isolation issue, unrelated; all pass standalone).

## Known deferrals (ledgered during review, need follow-up issues)

1. **Snapshot reconcile vs the same-cid splice**: after a replacement, the venue's order reports fills counted from zero; a snapshot that wins the venue-newer race can clobber the cumulative `filled_quantity` (pre-existing today for HL native modify). Needs a `filled_base` offset or venue-id-aware adoption — deliberately not attempted here (snapshot-semantics design pass).
2. **No confirm watcher for the replacement submit**: an async-lost replacement submit is only recovered by the periodic snapshot (`LocalOrderMissing → ResolveMissingOrder`). A meaningful watcher needs a new reconciler task type (the existing `AwaitOrderConfirm` would self-resolve on our own routed event).
3. **Backtester OME has no partial-fill matching** (fills are atomic), so the partial-fill replace race can't be simulated — covered by AM-level integration tests here and by a planned testnet e2e in the exchanges repo (`exchanges:docs/superpowers/plans/2026-08-05-cancel-fill-reporting-conformance.md`, includes the `venue_filled_quantity` conformance contract).

## Companion work (not in this PR)

- exchanges repo: conformance capability `cancel_reports_fills` + e2e partial-fill update scenario (spec committed there).
- quantkit: consumer-side hardening (corrupt-remaining guard, effective-position clamp, booked-ratio watchdog, CLOSED-state fixes) — `quantkit:docs/superpowers/plans/2026-08-05-maker-taker-venue-truth.md`.

### DIFF
--- a/docs/account-management/design.md
+++ b/docs/account-management/design.md
@@ -396,12 +396,15 @@ working order. Rather than adding a per-venue amend adapter to paper over the di
 `update_order` routes a partially-filled order through a framework-side cancel+replace: arm a
 `ReplaceIntent`, transition to `PENDING_UPDATE`, call `connector.cancel_order`. The connector
 never sees an amend call in this case — only the `cancel_order`/`submit_order` primitives every
-venue already has. The replacement resubmits under the **same client_order_id**, so a caller
+venue already has. The `filled == 0` check is inherently racy: a fill can land between the local
+check and the venue applying the amend, and the dialects then diverge on that one order after
+all. The window is bounded (one round trip) and self-corrects — the venue's amend ack carries
+the figures the next reconcile compares against. The replacement resubmits under the **same client_order_id**, so a caller
 watching one cid (e.g. a quantkit maker-taker slot) sees one continuously-live order, not a
 cancel+accept pair.
 
 **The intent lifecycle** (`ReplaceIntent` in `AccountState`, keyed by cid — `desired_remaining`,
-`price`, `filled_at_request`, `armed_at`, `filled_at_cancel`):
+`price`, `filled_at_request`, `armed_at`, `filled_at_cancel`, `canceled_venue_oid`):
 
 1. `update_order` arms the intent and fires `cancel_order`. A synchronous failure here (the
    cancel never reached the venue) clears the intent immediately and reverts via a synthetic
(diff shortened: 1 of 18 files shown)


yuriy-arabskyy:
Superseded by https://github.com/xLydianSoftware/Qubx/pull/368: review concluded the replace orchestration was over-engineered relative to real usage (the only production caller repriced only). The replacement switches update_order to a total-quantity contract with connector-owned amend dialects — see docs/superpowers/specs/2026-08-06-update-order-total-quantity-design.md on the new branch. Salvaged from this branch: the reducer never-negative-remaining regression, both sim reprice tests, and the design-doc analysis.

PR #375, opened 2026-08-12
fix(rate-limiting): enforce order pools, make gates loop-agnostic, charge real ccxt weights

bogdanKaftanatiy:
Addresses the CCXT rate-limiting audit in xLydianSoftware/exchanges#32.

## How it works now

Rate limiting has two layers that were previously conflated.

**Pools are the budgets.** Each venue declares an IP-scoped `ccxt_rest` pool and, where the venue
publishes one, an account-scoped `orders` pool, with capacity and refill taken from the venue's own
documentation. **`endpoint_map` is routing** — which pools an operation debits, not what the limits
are. The shape is shared across ccxt venues (`_default_endpoint_costs`) because only the numbers
differ; Kraken Futures overrides it, since that venue has one shared budget where an order costs 5x
a read and no separate order counter.

**REST weight** is charged by a throttle installed on the ccxt exchange. `install_rate_limiter_hooks`
replaces `exchange.throttle`, which is what takes ccxt's own throttler out of the REST path — `fetch2`
gates the call on `enableRateLimit`, so that flag stays on. ccxt supplies the per-endpoint cost, so a
1000-bar klines page costs 5 rather than 1. The same function installs the response-header hook, and
is now shared by `ExchangeManager` and `CcxtStorage`; previously only the former had it, so warmup and
aux readers ran unthrottled by us and double-throttled by ccxt.

**Order counts** are charged by the connector, which acquires `create_order` / `cancel_order` /
`edit_order` before the venue call. An order therefore debits two independent budgets: one IP weight
via the throttle, one order slot via the connector. A closed gate rejects before the request leaves,
as a typed event carrying `code="RateLimitGateTimeout"`.

**Venue feedback** arrives through response headers where the header's scope matches the pool it would
pin. For Binance that is `X-MBX-USED-WEIGHT-1M` for the IP pool and a window-matched
`X-MBX-ORDER-COUNT-*` for the order pool — the header name is derived from `capacity / refill_rate`,
so one parser serves `binance.um` (10s window) and `binance.pm` (1m window) even though both resolve
to the ccxt id `binanceusdm`. `-1` is Binance's "not applicable" sentinel and is ignored. Where no
header has the right scope, the model runs alone and 429s close the gate reactively.

**Concurrency.** One limiter instance is shared by the runner loop, the storage loop and each warmup
loop, which live on different threads. Gates are a monotonic deadline rather than an `asyncio.Event`,
and the token bucket takes its reservation under a `threading.Lock` and sleeps the deficit outside it.
Neither binds to an event loop, and a gate can be closed from a thread with no loop at all.

**Scoping.** Account-scoped pools derive their Redis scope id from a hash of the venue API key.
Previously they fell back to the literal `"local"`, which on a shared Redis backend would have
collapsed every bot onto one order budget.

## What changed

| Area | Change |
|---|---|
| `connectors/ccxt/connector.py` | order/cancel/edit acquire the `orders` pool; gate timeouts become typed rejects; venue order-limit errors close the orders gate |
| `connectors/ccxt/rate_limits.py` | `install_rate_limiter_hooks` extracted and shared; `_default_endpoint_costs`; venue configs corrected; window-matched order-count sync; case-insensitive header lookup; OKX parser removed |
| `connectors/ccxt/exchange_manager.py` | delegates to the shared installer; attach is idempotent |
| `data/storages/ccxt.py` | installs the shared hook before `load_markets`; per-page flat acquire and the best-effort `last_response_headers` sync removed |
| `rate_limiting/pools.py`, `utils/rate_limiter.py` | deadline gates, reservation bucket, oversized-weight clamp |
| `rate_limiting/manager.py`, `engine.py`, `utils/runner/runner.py` | account scope id; `rate_limit.gate_timeouts`; warning when rate limiting is disabled |
| `rate_limiting/redis_backend.py` | `acquire` returns the time it actually waited, so `rate_limit.wait_seconds` is meaningful under Redis |

### Venue configs

| Venue | `ccxt_rest` | `orders` | Note |
|---|---|---|---|
| binance (spot) | 6000 / 100 | 100 / 10 | orders capacity was 10 |
| binance.um | 2400 / 40 | 300 / 30 | models the 300/10s window |
| binance.cm | 2400 / 40 | 1200 / 20 | IP was 6000; dapi publishes only 1200/1m |
| binance.pm | 6000 / 100 | 1200 / 20 | previously fell through to the spot config |
| okx | 20 / 10 | 60 / 30 | |
| bybit | 120 / 24 | 10 / 10 | |
| kraken | 20 / 0.33 | 60 / 1.0 | |
| kraken.f | 125 / 12.5 (account) | — | previously got the Kraken *spot* config |

`binance.um.mm` and ccxt-id-shaped names (`binancepm`, `binanceqv_usdm`, `binance_um_mm`) now route
to their venue config instead of the generic fallback, which warns when it is used.

## Issue coverage

Fixed: #1 orders pool acquired, #2 binance.pm/cm configs, #3 loop-bound primitives, #6 disabled-state
logging and gate-timeout visibility, #8 storage weights and double-throttling, #9 dead auto-derivation,
#11 misleading docstrings, #12 tests.

Not built, with the reasoning recorded in the code: #4 (ccxt's per-api-group cost scaling is not a
uniform transform — Binance's papi table mixes IP-, order- and weight-denominated entries, so a single
scalar would be wrong in both directions; the unit is documented on `PoolConfig` and pinned by
`test_ccxt_cost_contract.py`), #5 (no endpoint signal exists at the throttle seam; the lock-across-sleep
that made this acute is gone), #7 (ccxt.pro already throttles subscribe frames per connection, which is
the scope the venue limits, and the seam runs after the handshake so it cannot affect connection rate),
#10 for Bybit and Kraken (Bybit's `X-Bapi-Limit-*` are per-endpoint and Kraken publishes none).

#13 belongs to xlydian-platform.

## Verification

2615 unit tests pass. New coverage includes cross-loop gate and bucket use, the reservation bucket's
sync semantics, throttle weight semantics end to end, header parsing and registry hygiene, the account
scope id, and a ccxt contract test pinning the cost numbers the design depends on.

Also exercised against Binance USD-M testnet through the production code path: real per-endpoint costs
charged, ccxt's throttler never invoked, IP-weight and order-count header sync landing on the pools,
the orders pool debited by real order flow, a closed gate rejecting without reaching the venue, one
limiter shared across two event loops on two threads, and the storage path charging real page weights.

### DIFF
--- a/docs/connectors/rate-limiting.md
+++ b/docs/connectors/rate-limiting.md
@@ -10,7 +10,7 @@ A **pool** represents one independent rate limit enforced by an exchange. Most e
 
 There are two pool types:
 
-- **Rate pool** (`pool_type="rate"`) — Time-based token bucket. Tokens refill at a constant rate. When depleted, a gate closes and reopens on a timer. Used for request weight limits, message rate limits, etc.
+- **Rate pool** (`pool_type="rate"`) — Time-based token bucket. Tokens refill at a constant rate. When the exchange reports a limit hit, a gate closes until a deadline. Used for request weight limits, message rate limits, etc.
 
 - **Quota pool** (`pool_type="quota"`) — Externally managed budget with no time-based refill. The exchange controls replenishment (e.g., Lighter's volume quota which is replenished by trading volume). When depleted, the gate closes and stays closed until the exchange reports positive remaining via `sync_from_exchange()`.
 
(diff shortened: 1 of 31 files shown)


bogdanKaftanatiy:
Review follow-ups pushed.

**Pool sizing** — the throttle override changed *what* gets charged (ccxt's real per-endpoint cost, not a flat 1), which invalidated capacities written as request counts. Only Binance had been re-derived:

| venue | was | now | effect |
|---|---|---|---|
| bybit | 120 / 24 | 100 / 100 | market data 4.8 → 20 req/s (was 25× stricter than the venue rule, and 2× stricter than the ccxt throttler it replaced) |
| okx | 20 / 10 | 4 / 8 | burst breached every endpoint budget by 2× |
| kraken spot | 20 / 0.33 | 10 / 1.0 | warmup 0.28 → 0.83 req/s; it modelled the *private* counter while OHLC/Trades/Depth are public and on a separate ~1 req/s IP rule |

**`gate_max_wait` 15s → 35s** — it was shorter than the cooldowns it has to wait out (30s), so one reported 429 timed out every waiter by construction.

**`MultiStorage` no longer latches** — it cached the reader set built while a storage was rate limited, dropping that storage for the whole run. Reachable from a single 429 in combination with the above.

**`RatePool.sync`** kept no reference to its background task and never retrieved its result, so a Redis blip logged an unhandled exception on every header sync.

**Disabled-rate-limiting banner** is INFO rather than WARNING — no `rate_limiting` section is the common case.

Each has a regression test; the gate and pool-window invariants are now pinned per venue. Re-verified against Binance USD-M testnet: a 30s gate recovers at 30.3s instead of timing out at 15s, and still times out at 35s on a longer cooldown.

yuriy-arabskyy (approved):

PR #389, opened 2026-08-22
fix(account): keep the client id the venue reported for external orders

dmarienko:
## What

An order classified `EXTERNAL` had its client id replaced with `ext:<venue_order_id>` — in both the snapshot-recovery path (`reconciler.py:497-500`) and the websocket materialization path (`reducer.py:149`).

On a venue that addresses cancel and modify **by client id**, that rewrite is fatal. `int("ext:…")` raises, the connector refuses the cancel locally, and it never reaches the venue. The order rests forever and the executor retries it every tick.

Classifying an order as EXTERNAL is a judgement about who placed it. The client id is a fact about how to reach it. The judgement must not overwrite the fact.

## Found live

A connector version change on a Lighter pool left resting orders whose client ids the new connector does not mint (different id scheme). 22 were classified EXTERNAL, lost their ids, and became permanently un-cancellable — **261 refused cancels in 55 seconds**, cleared only by rolling the connector back.

## Change

- `basics.py` — `external_client_id(client_order_id, venue_order_id)`: keep a reported client id, synthesise `ext:<venue_order_id>` only when there is none
- `reducer.py:149`, `reconciler.py:497-500` — call it instead of rewriting unconditionally

`origin` is unchanged: an external order still reads EXTERNAL.

No collision guard, deliberately: a cid the state already holds never reaches either site. The differ counts such a snapshot order as matched (`diffs.py:331`) and `_resolve` matches by cid before materializing, so the guard would be dead code.

## Other connectors

Both already do the right thing at their own layer and only synthesise when the venue gave nothing — `ccxt/utils.py:188` (`clientOrderId or ext:<id>`), `hyperliquid/translate.py:82,254` (`cloid or ext:<oid>`). This stops qubx from discarding what they produced. Neither is affected functionally: ccxt cancels by both ids, Hyperliquid by `oid`. Lighter is the only venue where the client id is the sole handle.

## Tests

933 core + 670 connector, all green. Three tests asserted the old rule and now assert the new one; one new test pins the no-client-id fallback.

## Live verification

11 resting orders planted on a live Lighter account — 10 carrying old-scheme client ids, 1 carrying an id the new scheme accepts (the control). Identical in market, price, size, side and id length; only the leading digits differ.

| | before | after |
|---|---:|---:|
| adoption | `'EXTERNAL' order ext:11540473910500379` | `'EXTERNAL' order 484410420` |
| refused cancels | 3,176 | **0** |
| cleared | 1 of 11 (the control only) | **11 of 11** |

Reproduction script: `test-strategies` `loe_tests/scripts/plant_old_style_orders.py` (`--place` / `--status` / `--cancel`).

### DIFF
--- a/src/qubx/core/account_manager/reconciler.py
+++ b/src/qubx/core/account_manager/reconciler.py
@@ -31,7 +31,14 @@ from qubx.core.account_manager.diffs import (
 )
 from qubx.core.account_manager.state import AccountState, VenueAccountFigures
 from qubx.core.account_manager.state_machine import can_transition
-from qubx.core.basics import EXTERNAL_CID_PREFIX, Instrument, Order, OrderOrigin, OrderStatus, Position
+from qubx.core.basics import (
+    Instrument,
+    Order,
+    OrderOrigin,
+    OrderStatus,
+    Position,
+    external_client_id,
+)
 from qubx.core.events import (
     AccountSnapshot,
     DealEvent,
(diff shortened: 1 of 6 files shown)


yuriy-arabskyy (approved):
Reviewed: fix is complete within qubx (no other cid rewrite sites), no-collision claim verified on both reconciler and reducer paths, ccxt/HPL unaffected. Approving.

Issue #388, opened 2026-08-21
Document: on_start does not fire in live after warmup, so per-instrument state must not be built only there


dmarienko:
Not a bug — the behaviour is deliberate — but it is worth writing down, because a strategy can only learn it from an outage. It cost `lighter.reversals` a stop on 2026-08-19 (quantkit#131).

## The asymmetry

`on_start` fires once per `_strategy_state`, and the warmup runner is handed the live context's state object (`runner.py:1095`), so `is_on_start_called` is already true when the live context checks it (`processing.py:495`).

| | universe at live start | `on_start` in live | per-instrument state the strategy built |
|---|---|---|---|
| no warmup | 15 picked + 1 held | runs | 16 — matches |
| with warmup | 15 picked + 1 held | **does not run** | 15 — one short |

Same strategy code, same config, different outcome. With warmup, the callback a strategy would naturally build per-instrument state in never fires against the universe that is actually traded.

## How the extra instrument gets there

Legitimately. `context.py:429-433` merges open venue positions into `_initial_instruments` — you cannot manage a position in an instrument you do not subscribe to — and `:455` installs them with `skip_callback=True`. A position held across a restart, in a symbol the refit no longer picks, is exactly this case, and the state resolver then closes it as intended.

## What to document

- `on_start` does **not** fire in the live context when warmup ran; it fired inside the warmup context, against the warmup universe.
- Therefore per-instrument state must be built lazily on first use, or rebuilt in `on_warmup_finished`, not only in `on_start`.
- `ctx.instruments` at live start can legitimately contain instruments the strategy never selected and was never notified about.

The strategy-side fix for the bot that hit this is quantkit#135: create the missing state on first sight and warn, plus a resync at `on_warmup_finished`.

## Adjacent, still open

A per-instrument exception counts toward the global ten-failure breaker (`processing.py:450`), so one unregistered symbol stopped trading in the other fifteen. That one may be worth changing on its own.

PR #402, opened 2026-08-27
feat(core): first-class boot phase — unconditional state resolution, set_fit_on_start, strict sync gate, boot-fit failure handling

### LINKED ISSUE #363: processing: state resolver never runs when warmup is disabled (hidden coupling)
## Problem

The state resolver is gated on the warmup sim having left position objects:

```python
# src/qubx/core/mixins/processing.py:502-504
if self._context.get_warmup_positions() or self._context.get_warmup_orders():
    self._handle_state_resolution()
```

and the runner captures a `Position` for **every** warmup instrument, including flat ones (`src/qubx/utils/runner/runner.py:1133-1134` — the filter is `v.quantity is not None`, not `!= 0`).

Net effect is binary and invisible:

- warmup enabled → resolver always runs
- warmup disabled → `_warmup_positions` stays `None` → `{}` → falsy → **resolver never runs, silently**

## Impact

The resolver hook is where strategies seed their domain state from live positions at boot (frab: pair-book seeding via `_state_resolver`, which ignores the warmup-sim arguments entirely and reads `ctx.get_positions()`). Disabling warmup — a plain config change — silently disables position-based state seeding with no warning. The coupling is undocumented and non-obvious: state resolution has no semantic dependency on warmup, yet is only triggered by it.

## Expected

State resolution against live positions should run at boot regardless of whether a warmup sim ran, or at minimum the framework should warn loudly when a registered state resolver is being skipped.

## Solution

To be added after discussion (tracked in Live Trading project).

yuriy-arabskyy:
Makes live-strategy boot a first-class phase with explicit guarantees, replacing the flag-check block in `_run_strategy_pipeline` with a `BootStateMachine`-driven sequence:

```
WAIT_READY → ON_START → RESOLVE → RESTORE → WARMUP_FINISHED → BOOT_FIT → TRADING
                                                                  ↓ (retries exhausted)
                                                               BLOCKED (self-heals on a later successful fit)
```

Design spec: `docs/specs/2026-08-27-boot-phase-design.md` (in this PR) — includes the sequencing diagrams, the four-case `on_fit` table, and rationale. Implementation plan: `docs/specs/2026-08-27-boot-phase-impl.md`. Strategy-author docs: `docs/trading/boot-lifecycle.md`.

Closes #363. Documents the #388 asymmetry (`on_start` fires in the warmup context).

## What changed

- **State resolution runs at every live boot** (#363) — the `get_warmup_positions() or get_warmup_orders()` gate is gone. `REDUCE_ONLY`/`SYNC_STATE` gain an empty-output guard: all-empty sim args ⇔ no warmup output → hold the live book + loud warning, never flatten. Default resolver (`REDUCE_ONLY`) install moved from `_run_warmup` into `_handle_state_resolution` (works with warmup off). New `StateResolver.HOLD`: cancel stale orders, keep positions, emit nothing — the partner of `set_fit_on_start` for buffered/tracker-reconciling strategies. Dead `use_limit_order` computation dropped from `SYNC_STATE`.
- **`initializer.set_fit_on_start(True)`** — opt-in guarantee that the first `on_fit` runs in the live context against live positions (the warmup-sim fit no longer counts). Replaces the `ctx.trigger_fit()`-in-`on_warmup_finished` boilerplate and kills the double-fit on venues where the warmup sim runs no fit. No-op without warmup. Identical on first start and restart.
- **Strict account-sync gate** — boot actions never run against phantom-zero positions. The `ACCOUNT_SYNC_TIMEOUT` fall-through is removed; 15s is now the alert threshold (`boot.account_sync_blocked` gauge + warning), and boot holds until the snapshot applies, self-healing when it lands. Simulation/paper unchanged.
- **Boot-fit failure handling** — live fits latch `is_on_fit_called` only on success (simulation keeps latch-in-finally, backtest behavior unchanged). Boot fit failure → 3 attempts, 60s apart → `BLOCKED` (`boot.fit_failed` gauge, no `on_event` delivery) → self-heals on any later successful fit (scheduled or `trigger_fit`). Works in both fit executor modes (`FitCommitData` gains an `error` field for the threaded path). `on_warmup_finished` failure latches as before + `boot.warmup_finished_failed` gauge. Framework raises on the boot-fit pass count as failed attempts (deadline armed, no hot loop) and stay loud.
- **Boot health gauges** — `boot.state`, `boot.account_sync_blocked`, `boot.fit_attempts`, `boot.fit_failed`, `boot.warmup_finished_failed`.

## ⚠️ Behavior changes for existing bots (operator-facing)

1. **No-warmup bots now run boot resolution.** Default → hold + warning, no trading action. Bots with a registered custom resolver: it now actually fires at boot (previously silently skipped — this is the #363 fix; frab's profile).
2. **A failed boot fit now blocks trading** (bounded retry → `BLOCKED` + health event) instead of silently trading unfitted.
3. **An unsynced account now blocks boot** (+ health event) instead of proceeding after the 15s timeout with phantom-zero positions.

Nothing else changes for existing bots: scheduled (post-boot) fits, simulation, and paper behavior are unchanged (parity pinned by tests).

## Noteworthy implementation decisions (deviations from the spec's literal pseudocode, reviewed and accepted)

- Fit-outcome reporting to the boot machine is **phase-gated** (`BOOT_FIT`/`BLOCKED`), not `not is_trading` — the literal spec code would have let a mid-boot `trigger_fit`/scheduled fit jump the machine to TRADING and skip `on_start`/resolution/restore. Regression-pinned by `test_mid_boot_scheduled_fit_does_not_skip_the_boot_steps`.
- The threaded fit path records **framework** raises (not just strategy raises) into `FitCommitData.error`, so a fit whose body never ran can't latch as fitted; propagation unchanged.

## Testing

- Full unit suite: **2919 passed, 5 skipped, 0 failed**; zero simulation-parity breaks (backtester suites untouched and green).
- New: `tests/qubx/core/boot_test.py` (machine), `tests/qubx/core/mixins/boot_pipeline_test.py` (boot sequence, resolution wiring, sync gate, four-case `set_fit_on_start` contract, failure/BLOCKED/self-heal), threaded-path tests in `fit_executor_test.py`.
- `uv run mkdocs build --strict` clean; ruff clean on all touched files.

## Follow-ups (not in this PR)

- BLOCKED-release hygiene: the compound-improbable latched-flag-at-exhaustion corner, and a log line for resolver signals that queue through a block and execute at self-heal.
- Repo chores: `just test` hardcodes `-n auto` (OOM footgun on arc runners); `just style-check` is red on main baseline.

https://claude.ai/code/session_0165zZzjTJTURiqEPksotMni

### DIFF
--- a/src/qubx/restarts/state_resolvers.py
+++ b/src/qubx/restarts/state_resolvers.py
@@ -20,6 +20,22 @@ def _no_warmup_output(
     return True
 
 
+def _cancel_all_orders(ctx: IStrategyContext) -> None:
+    orders = ctx.get_orders()
+    if not orders:
+        return
+    logger.info(f"Cancelling {len(orders)} live orders ...")
+    for order in orders.values():
+        try:
+            # - route by the id we actually hold: venue id when acked, else the client id
+            if order.venue_order_id:
+                ctx.cancel_order(order_id=order.venue_order_id)
+            else:
+                ctx.cancel_order(client_order_id=order.client_order_id)
+        except OrderNotFound:
+            logger.debug(f"Order {order.venue_order_id or order.client_order_id} already cancelled or doesn't exist")
+
+
 class StateResolver:
     """
     Collection of static methods for resolving position mismatches between
@@ -53,16 +69,7 @@ class StateResolver:
         The recommended partner of initializer.set_fit_on_start(True): let the first
         live fit reconcile positions through the strategy's own tracker.
         """
-        orders = ctx.get_orders()
-        if not orders:
-            return
-        logger.info(f"HOLD resolver: cancelling {len(orders)} live orders, keeping positions")
-        for order in orders.values():
-            oid = order.venue_order_id or order.client_order_id
-            try:
-                ctx.cancel_order(order_id=oid)
-            except OrderNotFound:
-                logger.debug(f"Order {oid} already cancelled or doesn't exist")
+        _cancel_all_orders(ctx)
 
     @staticmethod
     def REDUCE_ONLY(
@@ -136,15 +143,7 @@ class StateResolver:
             sim_active_targets (dict[Instrument, list[TargetPosition]]): Active targets from the simulation
         """
         # TODO: optimize with batch requests
-        orders = ctx.get_orders()
-        if orders:
-            logger.info(f"Cancelling {len(orders)} live orders ...")
-            for order in orders.values():
-                oid = order.venue_order_id or order.client_order_id
-                try:
-                    ctx.cancel_order(order_id=oid)
-                except OrderNotFound:
-                    logger.debug(f"Order {oid} already cancelled or doesn't exist")
+        _cancel_all_orders(ctx)
 
         # Get current live positions
         live_positions = ctx.get_positions()
--- a/src/qubx/restarts/state_resolvers.py
+++ b/src/qubx/restarts/state_resolvers.py
@@ -4,11 +4,29 @@ from qubx.core.exceptions import OrderNotFound
 from qubx.core.interfaces import IStrategyContext
 
 
+def _no_warmup_output(
+    sim_positions: dict[Instrument, Position],
+    sim_orders: dict[Instrument, list[Order]],
+    sim_active_targets: dict[Instrument, TargetPosition],
+) -> bool:
+    """All-empty sim args ⇔ no warmup sim ran (a sim that ran captures a Position per
+    instrument, flat included). Resolvers that steer toward sim state must hold then."""
+    if sim_positions or sim_orders or sim_active_targets:
+        return False
+    logger.warning(
+        "<yellow>State resolver received no warmup output — holding the live book as-is. "
+        "Register a custom resolver (or StateResolver.HOLD) to silence this.</yellow>"
+    )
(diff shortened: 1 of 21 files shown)


src/qubx/restarts/state_resolvers.py
```diff
@@ -23,6 +41,29 @@ def NONE(
...
+        live fit reconcile positions through the strategy's own tracker.
+        """
+        orders = ctx.get_orders()
+        if not orders:
+            return
+        logger.info(f"HOLD resolver: cancelling {len(orders)} live orders, keeping positions")
+        for order in orders.values():
+            oid = order.venue_order_id or order.client_order_id
+            try:
+                ctx.cancel_order(order_id=oid)
```
yuriy-arabskyy: you should cancel with parameter client_order_id if it's a client_order_id that you're using, otherwise order_id for venue_order_id

yuriy-arabskyy: Fixed in 04137294 — both HOLD and CLOSE_ALL now route through a shared _cancel_all_orders that passes order_id for acked orders and client_order_id otherwise (also removes the duplicated cancel loop).

PR #428, opened 2026-09-16
fix(core): an adopted explicit leverage set is read before the snapshot catches up

yuriy-arabskyy:
Seen on dev: editing Binance HEMIUSDT from 5x to 3x in the Instruments tab took ~3 minutes to show, though the venue accepted it within a second. For a **held** instrument `AccountManager.get_instrument_leverage` reads `Position.leverage`, which only refreshes on the next venue snapshot; the connector had already adopted 3x into its cache on the ack, but the account manager consults the connector only for flat instruments.

Rule now: while an instrument carries an explicit set (pinned) and the connector reports exactly that value, return it; otherwise snapshot-first as before. External changes made on the venue UI keep flowing in through the snapshot, and a refused set never reads as applied because the connector adopts only on success.

Two tests: adopted set read before the snapshot catches up; a not-yet-adopted set still reads the snapshot.

https://claude.ai/code/session_01TddMdRUUiYKzZASQRMLspc


yuriy-arabskyy:
Superseded: the connector will announce accepted/observed venue settings on the channel and the account manager will apply them to the Position, so every reader sees the new leverage within a second instead of a read-side special case.

PR #432, opened 2026-09-20
feat(data): Yahoo Finance storage with a Hive-partitioned parquet cache

dmarienko:
Adds a storage for daily and coarser bars of stocks, ETFs and indices, and the on-disk cache backend it needs.

## What is in it

**`ParquetCache(ICache)`** — the first persistent cache backend; `MemoryCache` was the only one. Layout is Hive-partitioned:

```
cache_key=<key>/data_id=<symbol>/data.parquet
_index/<key>.json      # covered ranges, globally and per symbol, plus each symbol's dtype
```

so DuckDB reads the cache with no export step and gets the partition keys as columns:

```sql
SELECT * FROM read_parquet('<root>/**/*.parquet', hive_partitioning = 1) WHERE data_id = 'SPY'
```

Range bookkeeping matches `MemoryCache`: a symbol counts as covered only when one merged range spans the whole request.

**`YahooStorage`** — builds its reader chain internally:

```
YahooReader(adjust) → CachedReader(ParquetCache) → YahooFetchReader(yfinance)
```

One copy per symbol is cached, holding the traded `open/high/low/close/volume` plus `adjclose`. `read(..., adjusted=True)` is the default and scales OHLC by `adjclose / close` **after** the cache, so the adjusted and raw views share one file instead of occupying two cache entries.

```python
storage = StorageRegistry.get("yahoo::~/data/yahoo/")
reader = storage["YAHOO", "STOCK"]
reader.read("SPY", "ohlc(1d)", "2010-01-01", "now").to_pd()             # adjusted
reader.read("SPY", "ohlc(1d)", "2010-01-01", "now", adjusted=False)     # as traded
```

Only `1d`, `1w` and `1M` are offered. Yahoo keeps roughly a month of 1-minute bars and two years of hourly ones, so an intraday request raises rather than quietly returning a short series.

## Why yfinance rather than a plain HTTP call

The intent was no new dependency, since `httpx` and `pyarrow` are already core. Measured from a clean session:

| | |
|---|---|
| `httpx` → `query1.finance.yahoo.com/v8/finance/chart/SPY` with browser headers | 429 |
| same on `query2` | 429 |
| `/v1/test/getcrumb` after a session cookie is established | 429 |
| `yfinance` | works |

Yahoo checks the TLS fingerprint, not only the headers. `yfinance` ships the browser impersonation that gets past it, so it goes in as the optional extra `qubx[yahoo]` with a lazy import and a clear error when absent. `YahooFetcher` is a separate class so another transport can replace it without touching the caching or the adjustment.

## Tests

`tests/qubx/data/test_yahoo_storage.py`, 15 offline plus 1 marked `integration`. No network in the offline set — the fetcher is stubbed, and the fixture encodes a 2-for-1 split so the adjustment is checked against a real discontinuity.

Covered: the vendor frame shape (MultiIndex columns, tz-aware index, missing `Adj Close`), rejection of intraday timeframes, cache roundtrip and persistence across instances, `covers`/`check`, window merging without duplicates, `clear` removing both the partition and its index, the adjusted and raw views, a second read inside a cached window not refetching, both views sharing one fetch, and DuckDB reading the Hive layout.

`224 passed` across `tests/qubx/data/`.

## Known limit

`ICache.check` is id-level, not range-level: a symbol cached for 2020-2024 and requested for 2020-2026 is refetched over the whole window rather than just the missing tail. Cheap for Yahoo daily, but it is not gap-filling. Making it so needs a range-aware `check` on `ICache`.

## Incidental fix

`MemoryCache._ranges_cover` used `re` as a loop variable, which shadowed the `re` module once `cache.py` imported it.


yuriy-arabskyy (approved):

dmarienko:
## Default cache path, no URI needed

Both storages now work without a path:

```python
StorageRegistry.get("yahoo")      # -> YahooStorage(~/.qubx/data/cached/yahoo)
StorageRegistry.get("dukascopy")  # -> DukascopyStorage(~/.qubx/data/cached/dukascopy)
```

`get_local_data_cache_folder(name)` in `qubx.utils.misc` builds on the existing `get_local_qubx_folder()` and returns `~/.qubx/data/cached/<name>`, creating it. Each storage gets its own subfolder deliberately — `ParquetCache` keys on data type and symbol only, so Yahoo's SPY and Dukascopy's SPY would otherwise land in the same file.

An explicit path still wins: `StorageRegistry.get("yahoo::/mnt/data/yahoo/")`.

Also fixed on the way: `qubx.data` imported `YahooStorage` but not `DukascopyStorage`, so the `@storage("dukascopy")` decorator never ran and `StorageRegistry.get("dukascopy")` raised `Unknown storage type` unless you had imported the module yourself. The tests did not catch it because they import the module directly; there is now a test that asserts both names are registered after a plain `import qubx.data`.

## Dukascopy rate limiting — it is the User-Agent, not a rate

Measured one header at a time, 0.5s spacing after a quiet period:

| headers sent | result |
|---|---|
| none | 429 at request 45 |
| Referer only | 429 at request 74 |
| Accept only | 429 at request 60 |
| **User-Agent only** | **200 requests, no 429** |
| all three | 600 requests, no 429 |

Referer and Accept sit inside the spread of unheadered runs (45, 51, 58, 65, 111). The feed throttles `Python-urllib/3.12`; a browser string runs clean. The datafeed fetcher now sends one. Referer is kept because the catalogue endpoint 403s without it.

The ceiling with an agent set is unmeasured — above 600 requests at 0.18 req/s, nothing tighter. `requests_per_second` is 5.0 as a conservative default rather than a measured limit, with the 429 handling as the backstop.

PR #433, opened 2026-09-20
docs: drop the In Production section from the README

dmarienko:
Removes the `## In Production` section. Cleanup only — no other change to the README, no code touched.


dmarienko:
Folded into #432 instead.

